PR #5, opened 2024-10-26
Setup/linting action

Darrow8:
I worked with Yiheng during the Saturday sats meeting to fix the linter github action.


devYaoYH:
LGTM. Thanks for the fix.

devYaoYH (approved):
LGTM. Thanks for the fix

PR #8, opened 2024-12-03
Add instructions for remote test bench access

devYaoYH:
Temporarily set up remote testing setup. Ping in #satellites-software slack channel for access.


niklas-vainio (approved):
Loog good!

PR #23, opened 2025-01-18
Core/command handler

Jackal-Studios:
Added a command handler with the following functionality:
1) parsing the radio queue with bytes in packets into commands
2) adding the parsed commands into their appropriate task queue
3) A command consists of a command id and data for that command, look into documentation for more info
4) multiple commands can be sent in one packet, big commands can be split into multiple packets and handled accordingly (assuming they come in the right order, if not there is a relatively simple fix), some packets can also be terminated early using a stop byte (see documentation).

Take a look at the documentation in the docs folder! 
- The satellite documentation describes how to add new commands (and has some Merging notes!)
- The ground documentation shows functions that can be implemented in the ground sending commands software for sending commands. It also offers some testing code that can be run on the pico to simulate the radio input from the ground.

If you have any questions contact:
- Sasha Luchyn GitHub: @Jackal-Studios
- Michael Dalva Github: @spirou7


shetaye:
I like the approach. However I think that handling commands that straddle the packet boundary is really complex because packets will inevitably arrive out-of-order, duplicated, or not at all. I hate to say it, but I think you should remove this functionality for now. From the looks of it you will simplify a lot.

shetaye (approved):

PR #28, opened 2025-01-28
Testing/refactor into modules

devYaoYH:
**PENDING HARDWARE TESTING of .uf2 BINARY**

**URGENCY: we should refactor into a more modular test-able structure and enforce some best-practices ASAP before folks start checking in various additional drivers and tasks. This WILL get exponentially harder to longer we leave it.**

Refactor codebase into modular structure. Better encapsulation of various drivers, states, tasks. Enables testing component-wise with unit tests (WIP but see print_task_test.c).

Some invariants:

- States should depend on some number of Tasks and maybe also include each other (next state).
- Tasks should not depend on states (inversion)
- Common header files should not have to depend on .c code (macro -> error is an issue here, but really the LOG_* macros should be in a separate logger library)

Best practices:
- The only header file one should include in .c is the corresponding .h file!
- `#pragma once` in all header files to avoid multiple symbol definitions.
- CMakeLists.txt per subdirectory (even if it's just a list of `add_subdirectory(...)`).
- run `cmake` from the project root (i.e. above `/src`).

Remaining issues with this structure:
- Actual states to be executed is defined as part of the scheduler (this should ideally be abstracted out from scheduler.c)
- rfm9x.h is included in slate.h for the definition of rfm9x_t and this propagates to a number of downstream dependencies requiring inclusion of `src/drivers/rfm9x` even though they don't really need it. Perhaps a note to rethink design of slate struct definitions.
- `extern` task/state definitions in the .h file (these are currently defined in the .c file, we *could move it, but I've left it as-is for now).


devYaoYH:
Update, testing via the remote setup, the built binary runs, but we get an issue with the radio driver:

```
user@raspberrypi:/home/common $ cat /dev/serial/by-id/usb-Raspberry_Pi_Pico_E6635C08CB47112E-if00 
[INFO]    main: Slate uses 128 bytes

[INFO]    main: Initializing everything...

[INFO]    sched: Enumerated 3 tasks

[INFO]    sched: Initializing task print

[INFO]    Test task is initializing...

[INFO]    sched: Initializing task blink

[INFO]    sched: Initializing task radio

[ERROR]   /Users/yaoyiheng/_Projects/SSSI/samwise-flight-software/src/drivers/rfm9x/rfm9x.c:655 Assertion failed: rfm9x_get_preamble_length(r) == 8
```

shetaye (approved):

PR #40, opened 2025-02-25
Add beacon task in new project structure.

devYaoYH:
Added beacon task to run every 30s on running_state.

Currently only sends:

1. Current state (which would just be running)
2. Time spent in state

Other minor fixes in this PR:

- Defined dummy pin (29) for SAMWISE_WATCHDOG_FEED_PIN in PICO build mode
- Moved packet.h up from command task to src/slate to remove dependency between tasks
- Swap back to using github native runners instead of configured self-hosted solution


shetaye:
Looks good. I think the following radio fields would be helpful:

```c
uint32_t rx_bytes;
uint32_t rx_packets;
uint32_t rx_backpressure_drops;
uint32_t rx_bad_packet_drops;
uint32_t tx_bytes;
uint32_t tx_packets;
```

However I am concerned they will take too many bytes. Should be fine for now, can always lower precision in the future.

Also, we can clean up structures like this one

```c
    tmp_data[pkt_len++] = slate->time_in_current_state_ms & 0x11111111;
    tmp_data[pkt_len++] = (slate->time_in_current_state_ms >> 8) & 0x11111111;
    tmp_data[pkt_len++] = (slate->time_in_current_state_ms >> 16) & 0x11111111;
    tmp_data[pkt_len++] = (slate->time_in_current_state_ms >> 24) & 0x11111111;
    tmp_data[pkt_len++] = (slate->time_in_current_state_ms >> 32) & 0x11111111;
    tmp_data[pkt_len++] = (slate->time_in_current_state_ms >> 40) & 0x11111111;
    tmp_data[pkt_len++] = (slate->time_in_current_state_ms >> 48) & 0x11111111;
    tmp_data[pkt_len++] = (slate->time_in_current_state_ms >> 56) & 0x11111111;
```

Using a `memcpy` and being careful about endianess. Since we are running on the same architecture for everything, let's just assume the wire endianess is the RP2350 endianness, so a `memcpy` should work fine.

(By the way, is there a reason you used `0x11111111`? I would have assumed `0b11111111`)

devYaoYH:
PTAL @shetaye I've implemented the memcpy suggestion and added the new radio task fields into the serialization.

shetaye (approved):
Looks great!

PR #62, opened 2025-05-10
Fix radio driver issue with integer division giving us a drift in Frequency of +0.06%

devYaoYH:
See snippet here: https://onlinegdb.com/IGBj2QNsp for verification.

Integer division used here: `_RH_RF95_FSTEP = 32000000 / 524288` causes a truncation from 61.035 to 61.

As a result, when we compute the frequency to set on the rfm9x chip, there is a +0.06% increase in the configured frequency. Given we used a target frequency of 438.1MHz, this resulted in a 263KHz increase in the sending frequency. This is around 2x the bandwidth configured (125KHz) and hence, we weren't able to pick up any signals (or even noise) during our testing.

Fixed this by reordering the operations and reducing our configured frequency value to be limited to 0.1MHz steps.

During flight mode, to reduce errors, a set of fixed constant values are used for 438.1MHz.


QuackWifHat (approved):
LGTM

PR #65, opened 2025-05-17
Write boot counter into slate and add to beacon telemetry packet

devYaoYH:
Number of times the picubed has rebooted as a unit32_t will be written as the 2nd field in beacon packet.


QuackWifHat (approved):
LGTM, 10/10

PR #67, opened 2025-05-17
feat: integrate TinyCrypt library for packet authentication

developStorm:
- Added TinyCrypt as a subdirectory in CMakeLists.txt.
- Created packet.c and packet.h for packet structure and authentication logic.
- Updated beacon and radio tasks to link with the packet library.
- Implemented packet authentication check in command task.

The entire lib/tinycrypt folder, except `lib/tinycrypt/CMakeLists.txt`, is from https://github.com/intel/tinycrypt and probably won't need be reviewed.


devYaoYH:
2 initial comments:

1. We can inject key during build time by defining CMAKE variables which is then visible as macros in the c code
2. Use boot counter as monotonically increasing sequence (see PR #65)

src/tasks/command/command_task.c
```diff
@@ -101,6 +101,13 @@ void command_task_dispatch(slate_t *slate)
 
         if (successful_peek)
         {
+            if (!is_packet_authenticated(&packet))
```
devYaoYH, line 104: Suggest you define a build flag so for convenience we can set at build time whether we want to enable packet authentication or not.

In CMakeLists.txt:
option(AUTH "Enable command packet authentication")

Then in here some version of:
#ifdef AUTH

So we can `cmake .. -DAUTH` during build time.

src/packet/packet.h
```diff
@@ -0,0 +1,45 @@
...
+typedef struct
+{
+
+} TASK1_DATA;
+
+typedef struct
+{
+    bool yes_no;
+    uint16_t number;
+} TASK2_DATA;
```
devYaoYH, line 45: Can we move these declarations into command_task.h? Feels out of place here since it's not used within packet.c

The only place I'm seeing this referenced is in command_task.c

src/packet/packet.c
```diff
@@ -0,0 +1,42 @@
...
+#include "logger.h"
+
+#include <tinycrypt/hmac.h>
+#include <tinycrypt/sha256.h>
+#include <tinycrypt/utils.h>
+
+#define PACKET_HMAC_PSK "StanfordSSIUMoJBTXpnXoZX3bQrNm5f"
+#define PACKET_HMAC_PSK_LEN 32
+
+_Static_assert(offsetof(packet_t, hmac) ==
```
devYaoYH, line 18: How does this `_Static_assert` work? (something pulled out from one of the tinycrypt files?

`hmac` variable here seems like it's floating undefined until line 31 (local function stack variable) unless I'm missing something here.

src/packet/packet.c
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+#include "packet.h"
+#include "logger.h"
+
+#include <tinycrypt/hmac.h>
+#include <tinycrypt/sha256.h>
+#include <tinycrypt/utils.h>
+
+#define PACKET_HMAC_PSK "StanfordSSIUMoJBTXpnXoZX3bQrNm5f"
```
devYaoYH, line 15: As previously mentioned in an earlier comment, have this be a CMake option definition so we can pass it in during build time instead of exposing it in our repo.

devYaoYH (changes requested):
Generally looks ok, minor nits in separate comments.

src/packet/packet.c
developStorm: fixed in 7bac89e748b56738e42d99497c35feceb72757f7

src/tasks/command/command_task.c
developStorm: fixed in 7bac89e748b56738e42d99497c35feceb72757f7

src/packet/packet.h
developStorm: fixed in 7bac89e748b56738e42d99497c35feceb72757f7

src/packet/packet.c
devYaoYH: Ah thanks! That's a useful check!

src/tasks/command/command_task.c
devYaoYH: Thanks!

Issue #64, opened 2025-05-17
Read charging current and voltage from MPPT board


devYaoYH:
MPPT Board document: https://docs.google.com/document/d/1o1tkhtAI3DVVbRXWPTltx64O2ZLYnPTUQDfJYijIO-Q/edit?tab=t.0

MPPT Board chip spec sheet: https://www.analog.com/media/en/technical-documentation/data-sheets/LT8491.pdf

MPPT AV Testing document: https://docs.google.com/document/d/1f7mS1cuhgpzzt6lWGUg5qEuXJnYqyc64XgwmhcSMzzA/edit?tab=t.0

Code snippet: 

```py
''' Sage Wu. Stanford University. May 13 2024 '''
''' This algorithm interfaces with the LT8491 MPPT over I2C, receiving important data such as battery charging status and temp '''
# LT8491 datasheet: https://www.analog.com/media/en/technical-documentation/data-sheets/LT8491.pdf
# pg 27 - 29 very useful for i2c stuff

# CircuitPython guide: https://docs.circuitpython.org/en/latest/shared-bindings/busio/

''' edited by Da-Hee Kim on 7/31/24 '''


import time
#import board
import busio
import board
# from bitstring import BitArray

# Define device address
LT8491_ADDR = 0x10
STATUS_REGISTER_ADDRESS = 0x12
# Configuration boolen
OUTPUT = True
CHARGE = False

# Tuple (immutable) of configuration register addresses and values
CFG = (
    ("CFG_RSENSE1", 0x28, 0x2710),     # CFG_RSENSE1
    ("CFG_RIMON_OUT", 0x2A, 0x0BC2),     # CFG_RIMON_OUT
    ("CFG_RSENSE2", 0x2C, 0x0CE4),     # CFG_RSENSE2
    ("CFG_RDACO",0x2E, 0x3854),     # CFG_RDACO
    ("CFG_RFBOT1", 0x30, 0x0604),     # CFG_RFBOT1
    ("CFG_RFBOUT2",0x32, 0x0942),     # CFG_RFBOUT2
    ("CFG_RDACI",0x34, 0x0728),     # CFG_RDACI
    ("RFBIN2", 0x36, 0x02DC),     # RFBIN2
    ("RDBIN1", 0x38, 0x03B9),     # RDBIN1
                )

# Tuple of telemetry register addresses
TELE = (
    0x00,               # TELE_TBAT: Battery temperature
[...]

PR #68, opened 2025-05-20
Add MPPT drivers and telemetry_task.c to poll each hardware component.

devYaoYH:
PR for #64.

MPPT driver added under `src/drivers/mppt`

Telemetry task currently polls from ADM & MPPT boards for power component status under `src/tasks/telemetry`


QuackWifHat (approved):
LGTM, pleaes resolve the conflict before merging please

devYaoYH:
There is potentially an issue with blocking i2c reads when device is detected but unresponsive. This looks to be causing some freezing of the software during testing. Will need to observe/debug it a bit more before I'm confident merging this.

src/tasks/telemetry/telemetry_task.c
```diff
@@ -0,0 +1,86 @@
+#include "telemetry_task.h"
+
+// Add power monitor instance
+static adm1176_t power_monitor;
+// Add MPPT instance
+static mppt_t solar_charger_monitor;
+
+void telemetry_task_init(slate_t *slate)
+{
+#ifndef PICO
```
bunny0514, line 10: Why do we need an ifndef for PICO? What does PICO contain and what conditions are we checking for when we use the ifndef?

src/tasks/telemetry/telemetry_task.c
```diff
@@ -0,0 +1,86 @@
...
+    LOG_INFO("Scanning the I2C bus...\n");
+    bool found_device = false;
+    for (uint8_t addr = 0x08; addr < 0x78; ++addr)
```
bunny0514, line 13: I feel like there may be a more efficient way to do this? Do we have the address of the devices saved anywhere? We could probably just directly reference them instead of testing whether we get a response from each address in the loop.

src/tasks/telemetry/telemetry_task.c
```diff
@@ -0,0 +1,86 @@
...
+                                false);
+        if (ret >= 0)
+        { // If ret is not an error code (i.e., ACK received)
+            LOG_INFO("Power Monitor Device found at 0x%02X\n", addr);
+            found_device = true;
+        }
+    }
+    if (!found_device)
+    {
+        LOG_ERROR("No I2C devices found.\n");
```
bunny0514, line 41: What happens to the workflow if we find no devices? Are we not able to proceed or do we just sort of skip over this part?

src/tasks/telemetry/telemetry_task.c
```diff
@@ -0,0 +1,86 @@
...
+
+    // Initialize mocked PICO MPPT
+    solar_charger_monitor = mppt_mk_mock();
+#endif
+}
+
+void telemetry_task_dispatch(slate_t *slate)
+{
+    // Read power monitor data from ADM1176
+    float voltage = adm1176_get_voltage(&power_monitor);
```
bunny0514, line 64: Do we have any precautions to make sure the values that adm1176_get_voltage and adm1176_get_current return are valid values and not garbage?

src/tasks/telemetry/telemetry_task.c
```diff
@@ -0,0 +1,86 @@
...
+    float current = adm1176_get_current(&power_monitor);
+    LOG_INFO("Power Monitor - Voltage: %.3fV, Current: %.3fA", voltage,
+             current);
+
+    // Convert float into mV and mA and write to slate
+    slate->battery_voltage = (uint16_t)(voltage * 1000); // Convert to mV
```
bunny0514, line 70: There might be edge cases where this wraps around weirdly due to the uint16_t. Should we clip these values based on known min and max voltage?

src/tasks/telemetry/telemetry_task.h
```diff
@@ -0,0 +1,15 @@
...
+#include "macros.h"
+#include "mppt.h"
+#include "pico/stdlib.h"
+#include "pico/unique_id.h"
+#include "pins.h"
+#include "rfm9x.h"
+#include "slate.h"
+#include "state_machine.h"
+
+extern sched_task_t telemetry_task;
```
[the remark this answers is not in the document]
devYaoYH: extern declares that there is this symbol for the compiler to link to, but does not define it.

The header file "telemetry_task.h" may be included by multiple other .c files so having it be defined in the header file itself will result in linker failures (to disambiguate which symbol definition to link to). The variable is defined in telemetry_task.c.

src/tasks/telemetry/telemetry_task.c
devYaoYH: This check if the build profile set is NOT PICO (which means one of the PICUBED builds). Primarily some hardware devices are present on the PICUBED board, but not the generic PICO.

Examples are MPPT (solar charging circuit), ADCS (altitude determination and control system), ADM (battery monitoring)...

So otherwise, I instantiate the mocked versions of these drivers for when we're building on the PICO platform.

src/tasks/telemetry/telemetry_task.c
devYaoYH: Yeah, if everything works well, the I2C addresses of these devices should be static (e.g. 0x10 and 0x29 if I remember correctly), but for testing, I'm scanning to find if the above assumed addresses are valid (due to hardware defects on the resister value that's setting these addresses).

src/tasks/telemetry/telemetry_task.c
devYaoYH: Good point, I think in the flight software, we should ignore their values if a failure case is detected. Even if there are no devices present at the I2C address, the default behavior of reading register values will return some random memory address values.

However, I think if some device exists but is in a bad state, our current version blocks indefinitely (see #75). We should use some timeouts instead of blocking forever.

src/tasks/telemetry/telemetry_task.c
devYaoYH: No such precautions, sanity checking values when we require them might be a good idea! Feel free to raise an issue on this :)

src/tasks/telemetry/telemetry_task.c
devYaoYH: Good callout, let me add some protections against this.

devYaoYH:
I'm going to merge this in first so we get a telemetry_task.c base as well as the utils.h file to work on (subsequent PRs pending on modifying these files).

PR #70, opened 2025-05-20
Drivers/payload uart merge

QuackWifHat:
Payload integration with main


devYaoYH (changes requested):
Run linter using `./format_all.sh` from repo root or `src/`.

src/drivers/payload_uart/payload_uart.c
devYaoYH, line 1: Perhaps it is possible to abstract some of the UART comms protocol into a separate lib?

So this can be used for ADCS communications as well.

QuackWifHat: From the best of my understanding, probably not because the protocol itself is specific to the packet structure coming in and out. That means ADCS would have to write their packets in a specific way that the driver expects it to be. The PiCubed uart protocol can be written to expect ADCS formed packets, however, that would also mean the payload would have to follow ADCS' packet format. There's also an issue with ACK and SYN timings, and many other things. So unless we already have a standardized way of how packets, ACKs, SYNs, etc. should be done via uart for all devices communicatin via uart, I'm not 100% sure if it would be worth our time to standardize this process because of the limited time we have. It might be easier to just have separate protocols that are similar that just works. 

ADCS also said that they're working on their own protocol so we don't have to necessarily handle it ourselves. (yay less work)

QuackWifHat:
Ran the ./format_all.sh, but when I ran git status it looks like there's nothing to really push. Not sure I'm understanding how to use the linter right

devYaoYH:
Looking into the logs for the failing lint check:

```
Posting comment(s)
  Notice: File src/slate/slate.h does not conform to Custom style guidelines. (lines 84, 85)
  Notice: File src/states/running/running_state.c does not conform to Custom style guidelines. (lines 17, 18, 19, 20)
  Notice: File src/states/running/running_state.h does not conform to Custom style guidelines. (lines 7)
  INFO:CPP Linter:3 clang-format-checks-failed
  INFO:CPP Linter:0 clang-tidy-checks-failed
  INFO:CPP Linter:3 checks-failed
```

So those lines above have formatting that doesn't conform to our style guides. `./format_all.sh` should run and modify these files in-place, so you should see 3 files changing afterwards.

Also make sure to pull this branch again (since I did the merge using the visual editor online - this commit won't be on your local).

devYaoYH:
Also please fix the failing builds for PICO-type platform builds.

You should be able to mock out these PICUBED pins:

```
/home/runner/work/samwise-flight-software/samwise-flight-software/src/drivers/payload_uart/payload_uart.c: In function 'payload_turn_on':
/home/runner/work/samwise-flight-software/samwise-flight-software/src/drivers/payload_uart/payload_uart.c:191:14: error: 'SAMWISE_RPI_ENAB' undeclared (first use in this function); did you mean 'SAMWISE_RF_SPI'?
  191 |     gpio_put(SAMWISE_RPI_ENAB, 1);
      |              ^~~~~~~~~~~~~~~~
      |              SAMWISE_RF_SPI
/home/runner/work/samwise-flight-software/samwise-flight-software/src/drivers/payload_uart/payload_uart.c:191:14: note: each undeclared identifier is reported only once for each function it appears in
/home/runner/work/samwise-flight-software/samwise-flight-software/src/drivers/payload_uart/payload_uart.c: In function 'payload_turn_off':
/home/runner/work/samwise-flight-software/samwise-flight-software/src/drivers/payload_uart/payload_uart.c:196:14: error: 'SAMWISE_RPI_ENAB' undeclared (first use in this function); did you mean 'SAMWISE_RF_SPI'?
  196 |     gpio_put(SAMWISE_RPI_ENAB, 0);
      |              ^~~~~~~~~~~~~~~~
      |              SAMWISE_RF_SPI
/home/runner/work/samwise-flight-software/samwise-flight-software/src/drivers/payload_uart/payload_uart.c: In function 'payload_uart_init':
/home/runner/work/samwise-flight-software/samwise-flight-software/src/drivers/payload_uart/payload_uart.c:208:23: error: 'SAMWISE_UART_TX' undeclared (first use in this function)
  208 |     gpio_set_function(SAMWISE_UART_TX,
      |                       ^~~~~~~~~~~~~~~
[ 58%] Building C object src/drivers/payload_uart/CMakeFiles/payload_uart.dir/__/__/__/pico-sdk/src/rp2_common/hardware_sync/sync.c.o
/home/runner/work/samwise-flight-software/samwise-flight-software/src/drivers/payload_uart/payload_uart.c:210:23: error: 'SAMWISE_UART_RX' undeclared (first use in this function)
  210 |     gpio_set_function(SAMWISE_UART_RX,
      |                       ^~~~~~~~~~~~~~~
/home/runner/work/samwise-flight-software/samwise-flight-software/src/drivers/payload_uart/payload_uart.c:214:15: error: 'SAMWISE_RPI_ENAB' undeclared (first use in this function); did you mean 'SAMWISE_RF_SPI'?
  214 |     gpio_init(SAMWISE_RPI_ENAB);
      |               ^~~~~~~~~~~~~~~~
      |               SAMWISE_RF_SPI
```

You can define these for non-PICO builds only with:

```c
#ifndef PICO
#define SAMWISE_UART_RX // as per normal pinout on PICUBED
#else
#define SAMWISE_UART_RX // as per mocked pins on PICO (pick something not currently in use)
#endif
```

src/drivers/payload_uart/payload_uart.c
devYaoYH: Ack. Thanks for checking. Let's go with separate implementations.

src/tasks/payload/payload_task.c
```diff
@@ -0,0 +1,100 @@
...
+    LOG_INFO("Payload task is initializing...");
+
+    LOG_INFO("Initializing UART...");
+    payload_uart_init(slate);
+
+    LOG_INFO("Turning on Payload...");
+    payload_turn_on(slate);
+
+    LOG_INFO("Waiting for Pi to boot up...");
+    sleep_ms(10000);
```
devYaoYH, line 20: For now ok... but think about how to avoid these long duration of sleeps as it'll block execution (and we only have a single thread).

Are there any I'm alive signals from the RPi we can scan/query for each time we try to establish a connection to it?

Also consider something like a stream/file-IO like interface with the RPi, open file --> read/write stuff --> close file. As **init for all tasks only run once immediately after boot**, and we'd like to be able to turn on payload only when we're needing it as a result of commands received from ground station (then turn it off on demand/after some timeout to conserve power).

devYaoYH: Just to be clear, you don't need to make changes to address this in the current PR, but keep these constraints in mind for a future fix.

src/drivers/payload_uart/payload_uart.c
```diff
@@ -0,0 +1,379 @@
+/**
+ * @author Niklas Vainio and Marc Reyes
+ * @date 2025-01-18
+ *
+ * UART packet and command handler routines for communicating with the RPi
```
devYaoYH, line 5: I am going to punt on reviewing this for dev velocity but let's circle back to this in case of performance constraints in the near future.

In particular, waiting on a syn when RPi might have crashed will block, and if we configure some static 1s timeout that'll also block other satellite operations in the meantime.

On first glance, ack doesn't appear to have pkt identifier information in this version of the UART protocol so we can't implement some queue pkt send --> return from task --> (come back and check in next iteration of task) if ack-ed kind of setup yet.

src/tasks/beacon/beacon_task.c
```diff
@@ -138,7 +138,7 @@ void beacon_task_dispatch(slate_t *slate)
 }
 
 sched_task_t beacon_task = {.name = "beacon",
-                            .dispatch_period_ms = 30000,
+                            .dispatch_period_ms = 6000,
```
devYaoYH, line 141: Can we keep this to 30s? Takes quite a bit of power to send out radio packets actually.

Funny thing: it crashes after this runs on USB power (I think due to insufficient power to transmit on radio antenna).

src/tasks/radio/radio_task.c
```diff
@@ -121,7 +123,7 @@ void radio_task_dispatch(slate_t *slate)
 }
 
 sched_task_t radio_task = {.name = "radio",
-                           .dispatch_period_ms = 100,
+                           .dispatch_period_ms = 5000,
```
devYaoYH, line 126: Also consider reverting this change, I think it's a no-op if the radio task runs and there are no packets to be transmitted in the tx_queue (so we can run it more frequently anyhow to catch pending packets added by other tasks).

But... I'm cognizant of the race-condition when we execute this and a packet transmission is already underway and we club the single FIFO queue on the rfm9x with another call to tx_done(). Any chance we could fix this with a simple 'locking' mechanism that writes a value into slate whenever we first enter tx_done and writes another value once tx_queue is empty and it switches back into receive mode (else branch starting line 34).

This is once again out-of-scope for your current PR, so consider rolling back this change for now?

QQ: do you know if queue_t in pico is multithread-safe? (atomic pop/push semantics)

devYaoYH (changes requested):
PTAL at the comments, some are forward-looking (and you don't have to address them in this PR), but primarily clean up some naming + (what I think might've been) spurious changes to the radio_task.

src/tasks/radio/radio_task.c
```diff
@@ -92,12 +93,13 @@ void radio_task_init(slate_t *slate)
...
-    rfm9x_set_tx_irq(&slate->radio, &tx_done);
+    // rfm9x_set_tx_irq(&slate->radio, &tx_done);
+    rfm9x_set_tx_irq(&slate->radio, 0);
     rfm9x_set_rx_irq(&slate->radio, &rx_done);
 
     // Switch to receive mode
-    rfm9x_listen(&slate->radio);
-
+    // rfm9x_listen(&slate->radio);
+    rfm9x_transmit(&slate->radio);
```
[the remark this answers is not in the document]
QuackWifHat: was added accidentally when testing radio during 433 fixing fiasco

src/tasks/beacon/beacon_task.c
QuackWifHat: was added accidentally when testing radio during 433 fixing fiasco

src/tasks/radio/radio_task.c
QuackWifHat: was added accidentally when testing radio during 433 fixing fiasco

src/common/pins.h
```diff
@@ -28,6 +28,10 @@
 #define SAMWISE_BURN_B_PWM_CHANNEL (PWM_CHAN_B)
 #define SAMWISE_BURN_RELAY (44)
 
+#define SAMWISE_RPI_ENAB (29)
+#define SAMWISE_UART_TX (0)
```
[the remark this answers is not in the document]
QuackWifHat: will fix right now

src/drivers/CMakeLists.txt
```diff
@@ -1,8 +1,10 @@
 add_subdirectory(adm1176)
 add_subdirectory(burn_wire)
 add_subdirectory(flash)
+add_subdirectory(payload_uart)
+add_subdirectory(payload_command)
```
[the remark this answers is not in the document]
QuackWifHat: alrighty

src/tasks/payload/payload_task.c
```diff
@@ -0,0 +1,100 @@
...
+        LOG_INFO("ACK:");
+        for (int i = 0; i < received_len; i++)
+        {
+            printf("%c", received[i]);
+        }
+        printf("\n");
+    }
+}
+
+void test_payload_command(slate_t *slate)
```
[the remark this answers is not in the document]
QuackWifHat: alright, will change it

src/drivers/payload_uart/payload_uart.c
QuackWifHat: The current model right now is that it sleeps for 100ms at most, so I don't think it should block too much the other satellite operations. The ACKs do not have some pkt identifier as it would just complicate things that currently work as is. We can implement it to make it more robust as you said, however, I believe that it can be handled later on when we have OTA. The payload and the other stuff related to it won't be turned on for long, so I don't think optimizing it completely is something we should prioritize for now. Rather I think many of the optimizations should come later when we have OTA implemented, unless these optimizations prove to be critical for satellite.

src/tasks/radio/radio_task.c
```diff
@@ -92,12 +93,13 @@ void radio_task_init(slate_t *slate)
     queue_init(&slate->rx_queue, sizeof(packet_t), RX_QUEUE_SIZE);
 
     // Install interrupt handlers
-    rfm9x_set_tx_irq(&slate->radio, &tx_done);
+    // rfm9x_set_tx_irq(&slate->radio, &tx_done);
+    rfm9x_set_tx_irq(&slate->radio, 0);
```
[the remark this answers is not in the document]
QuackWifHat: not sure, this was done when people were on my laptop testing the 433. I wouldn't look too deeply into it

src/drivers/CMakeLists.txt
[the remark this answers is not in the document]
devYaoYH: Will fix in a later PR.

src/drivers/payload_command/payload_command.c
```diff
@@ -0,0 +1,132 @@
+/**
+ * @author Marc Aaron Reyes
+ * @date 2025-05-21
+ *
+ * An easier way to send commands via UART to the Payload
```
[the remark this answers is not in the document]
devYaoYH: Got it, thanks for explaining the rationale here. Then I think drivers might not be the best place for this kind of helper library, since it's sort of like a logical wrapper around payload_uart and doesn't need to directly interface with hardware functions (one way to tell this is how it only required another driver header file, but nothing pico-sdk hardware specific headers).

This should go somewhere like a lib folder. But we can do that kind of cleanup later.

src/main.c
```diff
@@ -41,6 +41,7 @@ int main()
     /*
      * Initialize everything.
      */
+    sleep_ms(5000);
```
devYaoYH, line 44: this looks new... we probably don't need the explicit sleep here.

src/tasks/radio/radio_task.c
```diff
@@ -109,7 +111,7 @@ void radio_task_dispatch(slate_t *slate)
     // Switch to transmit mode if queue is not empty
     if (!queue_is_empty(&slate->tx_queue))
     {
-        rfm9x_transmit(&slate->radio);
+        LOG_INFO("Transmitting...");
```
devYaoYH, line 114: Note you're still removing the setting of radio mode to transmission here.

You can preserve the log statement, but let's add back the `rfm9x_transmit(&slate->radio)`.

PR #71, opened 2025-05-22
Add neopixel driver support on PICUBED.

devYaoYH:
neopixel RGB LED is attached to PICUBED on PIN 0.

Added driver contribution from Isaac from PR #25 from a while ago. Because file structure has diverged too much, easier time merging from a new branch off main.


QuackWifHat (approved):
LGTM

PR #82, opened 2025-05-25
Command Processing + Dispatch 

HaileThomas:


devYaoYH:
You should run `./format_all.sh` from within the repo root directory to format files as per our lint guidelines.

src/tasks/command/command_parser.c
```diff
@@ -0,0 +1,56 @@
...
+    {
+        /* Payload Commands */
+        case TAKE_PHOTO:
+        case DOWNLOAD_PHOTO:
+        case TAKE_AND_SEND_PHOTO:
+        {
+            PAYLOAD_COMMAND_DATA task;
+            strncpy(task.serialized_command, packet->data + 1,
+                    sizeof(task.serialized_command) - 1);
+            task.serialized_command[sizeof(task.serialized_command) - 1] = '\0';
```
devYaoYH, line 29: This always ensures that the final bytes of the copied-to string field is a null-terminator. **This feels more widely useful.** We should abstract it into some util function that does a 'safe-copy' of strings (in this way).

src/tasks/command/command_parser.c
```diff
@@ -0,0 +1,56 @@
...
+
+    Command command_id = (Command)packet->data[0];
+    LOG_INFO("Command ID Received: %i", command_id);
+
+    switch (command_id)
+    {
+        /* Payload Commands */
+        case TAKE_PHOTO:
+        case DOWNLOAD_PHOTO:
+        case TAKE_AND_SEND_PHOTO:
```
devYaoYH, line 24: It looks like you're directly passing through whatever serialized_command the command payload contains through to the task_data queue and (subsequently) into the RPi UART channel -- this is the expected behavior of the PAYLOAD_EXEC kind of command we discussed earlier.

Could we have this as just 1 PAYLOAD_EXEC command id?

src/slate/slate.h
```diff
@@ -54,16 +54,17 @@ typedef struct samwise_slate
     /*
      * Command switch
      */
-    queue_t task1_data; // queues of this kind will exist for each task called
-                        // from radio com
-    queue_t task2_data;
+    queue_t take_photo_task_data;
+    queue_t download_photo_task_data;
+    queue_t take_and_send_photo_task_data;
```
devYaoYH, line 59: What's the advantage of having separate task_data queues for payload commands (which all have the same data structure PAYLOAD_COMMAND_DATA)?

src/tasks/command/command_parser.c
```diff
@@ -0,0 +1,56 @@
...
+        {
+            PAYLOAD_COMMAND_DATA task;
+            strncpy(task.serialized_command, packet->data + 1,
+                    sizeof(task.serialized_command) - 1);
+            task.serialized_command[sizeof(task.serialized_command) - 1] = '\0';
+
+            if (command_id == TAKE_PHOTO)
+                queue_try_add(&slate->take_photo_task_data, &task);
```
devYaoYH, line 32: I was imagining TAKE_PHOTO passing a specific string of chars which specify some const string "take_photo,[],{}" as the PAYLOAD_COMMAND_DATA's task.serialized_command?

This currently just passes through whatever is received from the command packet's data field. Could we not consolidate this as the previously suggested single command PAYLOAD_EXEC?

src/tasks/command/command_parser.h
```diff
@@ -0,0 +1,45 @@
...
+
+#pragma once
+
+#include "packet.h"
+#include "slate.h"
+#include <stdbool.h>
+#include <stdint.h>
+#include <string.h>
+
+// Command IDs (must start from 1, as 0 indicates "not uploading")
```
devYaoYH, line 16: Hmm, what does "not uploading" mean? Could we not shift NO_OP up to the first and preserve TAKE_PHOTO as id value 1?

Generally, for style, either we assign values to all these enums, or not. But we should be consistent (and I'd prefer explicitly assigning values).

src/tasks/command/command_parser.c
```diff
@@ -0,0 +1,56 @@
...
+
+    switch (command_id)
+    {
+        /* Payload Commands */
+        case TAKE_PHOTO:
+        case DOWNLOAD_PHOTO:
+        case TAKE_AND_SEND_PHOTO:
+        {
+            PAYLOAD_COMMAND_DATA task;
+            strncpy(task.serialized_command, packet->data + 1,
```
devYaoYH, line 27: nit: use `packet->data[1]` instead?

.github/workflows/c_lint.yaml
```diff
@@ -6,6 +6,14 @@ on:
       - "**.c"
       - "**.cpp"
       - "**.h"
+  pull_request:
+    paths:
+      # only when a .c, .cpp, or .h file is changed
+      - "**.c"
+      - "**.cpp"
+      - "**.h"
+  workflow_dispatch:
```
developStorm, line 15: Could there be an issue in the Git history of this branch that caused this to appear as a new change? These are changes from #83 and should probably not show up here.

devYaoYH (changes requested):
Generally looks good! Thanks for refactoring it into command_parser and command_task!

devYaoYH:
Testing with NO-OP command works!

Screenshot 2025-05-26 at 5 30 53 PM

PR #97, opened 2025-05-27
Initialize mock ground station code with simple command sending and beacon receiving loop.

devYaoYH:
Mock ground station setup using Pi Pico + Adafruit rfm9x LoRA on a breadboard.


devYaoYH:
Force pushing this in, doesn't touch sats code.

PR #105, opened 2025-05-31
Refactored actual payload task execution into the payload_task

devYaoYH:
Command task's responsibility should be to accept, parse and delegate commands into separate tasks.

payload_task should handle the actual execution of the payload commands received.

Currently it attempts to retry previously issued commands up to 3 times before discarding it.

If the commands do succeed, up to 3 commands are executed per task.


QuackWifHat (approved):
LGTM, 10/10 engineering.

PR #113, opened 2025-06-05
Drivers/fixed solar

devYaoYH:
Added support for reading from fixed solar panel charging status and fault


devYaoYH:
This should be good to go. Only adds GPIO inits and single pin status reads HIGH/LOW with boolean helper functions.

PR #120, opened 2025-06-11
Fixed crc32 calculation overflow bug

QuackWifHat:
Bug:
When calculating the crc32 for the payload, the for loop induces a overflow which results in an infinite loop.

Fix: 
Cast for loop variable as an int rather than a size_t which is unsigned.

### DIFF
--- a/src/drivers/payload_uart/payload_uart.c
+++ b/src/drivers/payload_uart/payload_uart.c
@@ -57,7 +57,8 @@ static void uart_rx_callback()
 // (not using fast version because we want small binary size)
 unsigned int crc32(const uint8_t *message, uint16_t len)
 {
-    size_t i, j;
+    size_t i;
+    int j;
     unsigned int byte, crc, mask;
 
     i = 0;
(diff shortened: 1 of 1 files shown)


src/drivers/payload_uart/payload_uart.c
```diff
@@ -57,7 +57,8 @@ static void uart_rx_callback()
 // (not using fast version because we want small binary size)
 unsigned int crc32(const uint8_t *message, uint16_t len)
 {
-    size_t i, j;
+    size_t i;
+    int j;
     unsigned int byte, crc, mask;
```
devYaoYH, line 62: crc is not initialized here. Later on, you do `crc = crc ^ byte` which may result in different outcome than expected.

Should crc start from 0 each time?

devYaoYH (changes requested):
The variable `j` used in lines 70-74 for loop appears to be local, if all you need to do is to loop 8 times (and you do not reference the looping var within the for-loop), why not declare it normally as a local variable and loop upwards?

```
for (size_t j = 0; j < 8; ++j)
{
...
}
```

src/drivers/payload_uart/payload_uart.c
QuackWifHat: Unless I'm not understanding, we set crc as 0xFFFFFFFF on line 65. So no, we should not start crc off as 0 each time.

devYaoYH: Oh I see, I missed that! Thanks for pointing that out.

devYaoYH (approved):

PR #121, opened 2025-07-05
Payload uart timeouts

QuackWifHat:
Hello Yiheng, if you approve this you will get good luck for the next 1000 years.


devYaoYH:
Nice, if I'm understanding this correctly, instead of the RPi payload turning on by default, we only initialize the UART channel and wait for an explicit command to turn it on?

devYaoYH (approved):

PR #128, opened 2025-07-13
Picubed <- ADCS interface

niklas-vainio:
Remaining TODOs for software to decide/figure out:

- Powering on the board at init time (setting the `ADCS_EN` pin appropriately, confirm with Sage for whether it should be high/low)
- Dealing with a dead ADCS board - ideally if we haven't received any packets after a set amount of time (the ADCS board should send one every second or so), we should power cycle ADCS because something is propably wrong (we are effectively acting like an external watchdog)
- Packing the adcs telemetry into the actual telemetry to be downlinked (we don't need to send all of it if too big - the most important ones are the quaternion, state and boot count


src/common/pins.h
```diff
@@ -93,6 +93,11 @@
 #define SAMWISE_UART_TX (31)
 #define SAMWISE_UART_RX (30)
 
+#define SAMWISE_UART_TX_TO_ADCS (24)
+#define SAMWISE_UART_RX_FROM_ADCS (25)
+#define SAMWISE_ADCS_EN (26)
+#define SAMWISE_ADCS_UART (uart1)
+
```
devYaoYH, line 100: for successful build checks, we will need mocked pins in this file for the PICO board as well (although we won't use nor read from them).

src/states/running/CMakeLists.txt
```diff
@@ -13,6 +13,7 @@ target_link_libraries(running_state PUBLIC
   telemetry_task
   payload_task
   diagnostics_task
+  adcs_task
```
devYaoYH, line 16: let's try to keep this list sorted (I did a bad job here setting a poor precedence)

src/tasks/beacon/beacon_task.c
```diff
@@ -70,8 +72,14 @@ size_t serialize_slate(slate_t *slate, uint8_t *data)
                           .solar_current = slate->solar_current,
                           .device_status = get_device_status(slate)};
 
-    memcpy(data + name_len + 1, &stats, sizeof(beacon_stats));
-    return name_len + 2 + sizeof(beacon_stats);
+    // 2 Extra bytes: 1 for length of string, 1 for \0 terminator
+    memcpy(data + name_len + 2, &stats, sizeof(beacon_stats));
```
devYaoYH, line 76: this was a bug I introduced, it should still be `data + name_len + 1`. The length of string field was removed from beacon packing (we read until \0 for Kaitai decoding).

src/tasks/beacon/beacon_task.c
```diff
@@ -70,8 +72,14 @@ size_t serialize_slate(slate_t *slate, uint8_t *data)
...
+
+    // Copy adcs packet - device status will indicate if this is invalid
+    memcpy(data + name_len + 2 + sizeof(beacon_stats), &slate->adcs_telemetry,
```
devYaoYH, line 79: this starting byte addr correspondingly needs to be updated too

devYaoYH (changes requested):
@claude could you help take a look at the requested changes and modify this PR accordingly?

devYaoYH (approved):
Thanks claude (and @niklas-vainio)!

Issue #84, opened 2025-05-25
[Command] Add PING command which triggers response radio packet from satellite -> ground station


devYaoYH:
Upon receiving a PING packet, satellite should queue and send down a radio packet with the number of commands received **including this current PING command.**

PR #129, opened 2025-07-18
Feature/ping command

QuackWifHat:
Topic: PING command
Issue: #84 

Description:
Implemented PING command that loads a info packet onto the radio to be sent down as a beacon to the ground station. The packet contains information of commands that were processed, meaning every single command that was received by the command parser; not just commands that were executed properly. 

TODOs:
This command has not been tested yet with a mock ground station. but given how small of a feature this is putting a PR would be good for visibility and would be easier to fix later on anyway (time crunch is sad).


src/tasks/command/command_parser.c
```diff
@@ -40,10 +42,18 @@ void dispatch_command(slate_t *slate, packet_t *packet)
...
         }
-        case NO_OP:
+        case PING:
         {
-            LOG_INFO("Number of Commands Executed: %d",
+            LOG_INFO("Retrieving number of commands executed...");
+            char buf[MAX_PACKET_SIZE];
+
+            // Package interger value into a string
+            snprintf(buf, sizeof(buf), "Number commands executed: %d",
```
devYaoYH, line 51: I'm not entirely sure if this is the way we're adding packets into the slate. tx_queue expects some `packet_t` struct

devYaoYH (changes requested):

src/tasks/command/command_parser.c
QuackWifHat: you're right, let me fix that really quickly

src/tasks/command/command_parser.c
```diff
@@ -45,15 +45,31 @@ void dispatch_command(slate_t *slate, packet_t *packet)
...
                      slate->number_commands_processed);
 
+            // Create the packet
+            packet_t pkt;
+            pkt.src = 0;
+            pkt.dst = 0;
+            pkt.flags = 0;
+            pkt.seq = 0;
+            pkt.len = sizeof(data);
+            pkt.data = data;
```
devYaoYH, line 61: If you could help with some refactoring that would be great, ideally we'd want this response or any radio packet creation to be standardized and in a helper function somewhere, say within the drivers/rfm9x lib perhaps?

devYaoYH: Or actually just a util function within packet/packet.h would work as well. Like:

```c
void create_command_response_packet(packet_t *pkt, char *data) {
    ...
    // Helps to auto set pkt.src = 0, pkt.dst = 0, ... (but note that we'll likely have different values for these soon!)
}
```

QuackWifHat: The beacon task already does something like that, when I do this refactoring and create a standardized fuunction in drivers/rfm9x should I also refactor the beacon task too or leave it alone for now?

devYaoYH: Yup! That's the idea! We should have it in one place, so if you can move that too, that'll be great. But yeah, feel free to stop me from adding onto your place and let's raise a separate issue -> check this in first? I kinda want to test if the PING works.

devYaoYH (approved):
Thanks!

devYaoYH (changes requested):
Wait... I just saw you were using a different const for pkt data size, let's use PACKET_DATA_SIZE from packet.h instead! standardize to that

PR #132, opened 2025-07-19
Added manual override state into slate instead of an extern var

devYaoYH:


devYaoYH:
FYI @HaileThomas `strcmp == 0` when strings are equal -.-

PR #139, opened 2025-08-03
Adcs/add task to running state

devYaoYH:
Also fixed up the UART comms, there is a buffer somewhere in the UART hardware and we in fact do queue up bytes even without reading them immediately (it looks to live for up to 1s and maybe beyond that too).

### DIFF
--- a/ground_station/code.py
+++ b/ground_station/code.py
@@ -236,7 +236,9 @@ def decode_beacon_data(data):
         print("DEBUG: stats_start = {}, remaining bytes = {}".format(stats_start, len(data) - stats_start))
         
         # New beacon_stats struct size: 4+8+4+4+4+4+4+4+2+2+2+2+1 = 45 bytes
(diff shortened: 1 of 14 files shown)


devYaoYH:
FYI @niklas-vainio UART has some built-in buffering, pg 976 of the [RP2350 datasheet](https://datasheets.raspberrypi.com/rp2350/rp2350-datasheet.pdf), info on the actual size of buffer is sparse, some report 32Bytes. So if we keep comms packets between ADCS & PiCubed within this size window, we should be able to do without interrupts on either end!

devYaoYH:
@QuackWifHat assigning you this PR for review, perhaps useful to consider for the RPi side of things too? For the PiCubed, we use 2 different UART lines, uart0 for RPi and uart1 for ADCS, I'm going to assume that they have separate hardware FIFOs for buffering bytes along each line.

lundeen06 (approved):
looks good! still figuring out what telemetry we want to send, but this can be in a different PR

PR #144, opened 2025-08-17
Update claude.yml

devYaoYH:
Allowlist more contributors to using claude github action.


devYaoYH:
@claude could you allowlist 'tahmidjamal12' as well?

PR #153, opened 2025-10-02
Migrated build chain to use VSCode RPi extension.

devYaoYH:


devYaoYH:
@claude help me fix the cmake file to be compatible with VSCode RPi extension

PR #156, opened 2025-10-07
Update installation paths in ONBOARDING.md

megargayu:
/usr/opt does not exist on linux machines - instead, use /usr/local (this was tested on WSL2 & MingW32, so maybe it does on some linux machines? But /usr/local should work for everyone, so this change should still be fine.)


devYaoYH (approved):
LGTM, thanks for the update!

PR #168, opened 2025-10-15
Added ADCS mram driver to PiCubed

megargayu:
Need to merge the new mram driver code into the main branch to run tests properly & start work on the filesystem.


src/drivers/mram/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+add_library(mram mram.c)
```
devYaoYH: to compile this code, remember to add the subdirectory into the parent directories' CMakeList.txt

devYaoYH (approved):

PR #183, opened 2025-10-25
Added reading of single hardware pins to telemetry_task.c

microbites0:
Updated hardware pin readings on telemetry task & definitions in init.c


devYaoYH:
@microbites0 is this ready for review yet?

PR #189, opened 2025-10-29
Created heartbeat_check function in payload_task.c

sangeetsatpathy:
Implemented the heartbeat_check() function in payload_task.c
Sends a ping through the ping_command_test() function.
Checks the RPI_ENAB pin and does casework;

If we get a ping back, we know the RasPi is operational. The idea was that the RPI_ENAB pin should only be on if the payload_turn_off() function was called (it should not have been randomly turned off). So if the RPI_ENAB pin is off, check if the slate "is_payload_on" is True, which indicates if the payload_turn_off() function was called.

More details about each case is written within the code Logs.


QuackWifHat (changes requested):
Let it pass linter, and also needs to be part of the logic outlined in the issue ticket and what was discussed during Tuesday

src/tasks/payload/payload_task.c
```diff
@@ -135,7 +135,39 @@ void beacon_down_command_test(slate_t *slate)
     }
 }
 
-void ping_command_test(slate_t *slate)
+bool heartbeat_check(slate_t *slate)
+{
+    const uint64_t TIME_THRESHOLD = 1000;
```
devYaoYH: Also put this as a #define statement at the top of the header file. `#define PAYLOAD_HEARTBEAT_TIMEOUT_MS 10000` <-- We can do say 10 second first?

Include the unit of time measurement in the variable name, it'll be helpful to avoid mismatching later on.

src/tasks/payload/payload_task.c
```diff
@@ -135,7 +135,39 @@ void beacon_down_command_test(slate_t *slate)
...
+    if (slate->is_payload_on)
+    {
+        bool response_received = ping_command_test(slate);
+        uint64_t timeDelta = absolute_time_diff_us(
```
devYaoYH: Here the unit of measurement is microseconds? We should make sure we're not off by some order of magnitudes.

src/tasks/payload/payload_task.c
```diff
@@ -135,7 +135,39 @@ void beacon_down_command_test(slate_t *slate)
...
+            slate->payload_most_recent_ping_time, get_absolute_time());
+        if (timeDelta >= TIME_THRESHOLD)
+        {
+            payload_turn_off(slate);
```
devYaoYH: We may need to pivot on how we're restarting the Raspberry pi slightly. Context: https://ssi-teams.slack.com/archives/C2L29KW8J/p1747723117424069

src/tasks/payload/payload_task.c
```diff
@@ -30,6 +30,65 @@ void payload_task_init(slate_t *slate)
     // NOTE: Turning on payload is handled by command_parser
 }
 
+bool ping_command(slate_t *slate)
+{
+    char packet[] = "[\"ping\", [], {}]";
+    int len = sizeof(packet) - 1;
+    payload_uart_write_packet(slate, packet, len, 999);
```
devYaoYH: We can construct a new function `bool send_payload_exec(string)` function that handles:

sending a well-formatted string of payload into the payload through UART. This bit of logic is duplicated in the `try_execute_payload_command` innermost if statement.

Note that some changes will also be required for the try_execute_payload_command to keep its previous behavior consistent (i.e. when send_payload_exec returns true we [remove the command from the queue](https://github.com/stanford-ssi/samwise-flight-software/blob/e0e8dc18b07ecae86202d7e5930694bb0f33a67a/src/tasks/payload/payload_task.c#L56)).

This will replace this function here and wrap it with error checking. If the sending wasn't successful, we should just return false at this point without waiting for the receiving.

src/tasks/payload/payload_task.c
```diff
@@ -30,6 +30,65 @@ void payload_task_init(slate_t *slate)
...
+
+    safe_sleep_ms(1000);
```
devYaoYH: 1 second wait for each ping? This feels a bit long. @QuackWifHat what's a more reasonable time for this?

QuackWifHat: Also the execution for the payload side is also in the same order, which is why I think we don't need any waits here, baud rate is also 1152000 so it's fast enough to not need a wait for probably a few bytes at most

devYaoYH:
Synced with Marc, we won't need to use `seq_num`, let's just leave a dummy value and come around to clean that up later.

@sangeetsatpathy you were pretty close with the last pass of this PR, shall we try to get it in before Tuesday (for testing)?

src/drivers/payload_uart/payload_uart.c
```diff
@@ -417,7 +417,7 @@ uint16_t payload_uart_read_packet(slate_t *slate, uint8_t *packet)
     }
 
     // Read actual packet
-    bytes_received = receive_into(slate, packet, header.length, 1000);
+    bytes_received = receive_into(slate, packet, header.length, 50);
```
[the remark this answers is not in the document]
devYaoYH: Minor change, quick ping on this.

devYaoYH (approved):

PR #190, opened 2025-10-29
Made ground station code compatible with rpi pins

devYaoYH:
RPI detection is the fall-through from PICO2 & FEATHER M4.

Import default python libs when we don't have adafruit-circuitpython libs (on Raspberry Pi).

Raspberry Pi also doesn't have a default LED, so gate led operations behind a `LED_PIN is not None` conditional.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,6 @@
 **/.vite
 **/.DS_Store
 **/.vite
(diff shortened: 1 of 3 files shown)


QuackWifHat (approved):
LGTM

PR #201, opened 2025-11-01
 replaced strlcpy for compatibility 

megargayu:
strlcpy is an inefficient command that also doesn't work with some compilers (e.g. gcc on WSL Linux) because it relies on OpenBSD to be linked. This replaces strlcpy with https://nrk.neocities.org/articles/not-a-fan-of-strlcpy so that we have better compatibility throughout.

### DIFF
--- a/ground_station/code.py
+++ b/ground_station/code.py
@@ -1,9 +1,7 @@
+import adafruit_rfm9x
 import board
 import busio
 import digitalio
-import time
-import random
-import struct
 try:
     from adafruit_hashlib import sha256
 except:
@@ -12,8 +10,18 @@ try:
     import circuitpython_hmac as hmac
 except:
     import hmac
+import random
+import struct
 import sys
-import adafruit_rfm9x
+import time
+
+try:
+    # This lib might not be available on the CircuitPython platform.
+    import functools
+    # Override all print statements to immediately flush write buffer.
+    print = functools.partial(print, flush=True)
+except Exception as e:
+    print("functools lib not available: ", e)
 
 
 def detect_board():


src/tasks/beacon/beacon_task.c
```diff
@@ -72,8 +73,7 @@ size_t serialize_slate(slate_t *slate, uint8_t *data)
     LOG_INFO("State name: %s", slate->current_state->name);
     // Copy null-terminated name to buffer (up to MAX_STR_LENGTH - 1)
     size_t name_len = strnlen(slate->current_state->name, MAX_STR_LENGTH);
-    strncpy((char *)data, slate->current_state->name, name_len);
-    data[name_len] = '\0'; // Ensure null termination
+    strcpy_trunc((char *)data, slate->current_state->name, name_len + 1);
```
devYaoYH: Why the `name_len + 1`?

devYaoYH (changes requested):

src/tasks/command/command_parser.c
```diff
@@ -36,10 +36,8 @@ void dispatch_command(slate_t *slate, packet_t *packet)
         case PAYLOAD_EXEC:
         {
             PAYLOAD_COMMAND_DATA payload_command;
-            strncpy(payload_command.serialized_command, command_payload,
-                    command_payload_data_size);
-            payload_command.serialized_command[command_payload_data_size - 1] =
-                '\0';
+            strcpy_trunc(payload_command.serialized_command, command_payload,
```
megargayu: @devYaoYH I'm not sure if command_payload_data_size is the correct max size - I think so, right? Is there an off-by-one error here?

devYaoYH: I think this is correct, command_payload pointer points to the `&packet->data[1]` address instead (so we need the -1 that is also captured by the command_payload_data_size).

devYaoYH (commented):
LGTM, thanks!

megargayu:
@devYaoYH Can you mark the requested change as satisfied?

devYaoYH (approved):
Thanks for the changes!

Issue #175, opened 2025-10-22
[Ground Station] Add capability to read boot count and message id from satellite beacon to compute HMAC


devYaoYH:
Previous issue #96 

We need to dynamically update the boot_count and msg_id to recompute HMAC when the satellite restarts.

megargayu:
Closed by #203

PR #203, opened 2025-11-02
Read Boot Count and Message ID from Satellite Beacon

joeytene:
Addresses #175

WIP - Untested


devYaoYH (approved):
Approving, minor changes, unblock downstream migration in progress. This should not change the correctness of decoding telemetry packets when encryption is off.

PR #210, opened 2025-11-08
fixed snprintf use in PING command

### LINKED ISSUE #205: Replace `snprintf` in PING command
This is the current PING command implementation in command_parser.c:
```c
        case PING:
        {
            LOG_INFO("Retrieving number of commands executed...");
            uint8_t data[PACKET_DATA_SIZE];

            // Package interger value into a string
            int len =
                snprintf(data, sizeof(data), "Number commands executed: %d",
                         slate->number_commands_processed);

            // Create the packet
            packet_t pkt;
            rfm9x_format_packet(&pkt, 0, 0, 0, 0, len, &data[0]);

            // Add to transmit buffer
            LOG_INFO("Sending to radio transmit queue...");
            if (queue_try_add(&slate->tx_queue, &pkt))
            {
                LOG_INFO("Ping info was sent...");
            }
            else
            {
                LOG_ERROR("Ping info failed to send...");
            }
            break;
        }
```

The `snprintf` call is incorrect and needs to be replaced with `strncpy` or a similar function, as snprintf returns the length of the entire string, not the truncated length. (https://www.geeksforgeeks.org/c/snprintf-c-library/).

megargayu:
Fixes #205 by calculating the correct length after the snprintf command

### DIFF
--- a/src/utils/str_utils.h
+++ b/src/utils/str_utils.h
@@ -43,3 +43,19 @@ inline static int snprintf_len(char *dest, size_t data_size, const char *fmt,
         return (int)data_size;
     }
 }
+
+/**
+ * Replacement for strlcpy that works for static-sized buffers (dst) that
+ * don't care about truncation and just want to ensure null-termination.
+ * Adapted from https://nrk.neocities.org/articles/not-a-fan-of-strlcpy.
+ *
+ * @param dst Destination buffer.
+ * @param src Source string.
+ * @param size Maximum size of the destination buffer.
+ */
+inline static void strcpy_trunc(char *dst, const char *src, size_t max_size)
+{
+    if (memccpy(dst, src, '\0', max_size) == NULL)
+        dst[max_size - 1] =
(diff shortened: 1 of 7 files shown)


src/utils/str_utils.h
```diff
@@ -0,0 +1,45 @@
...
+ *       ensuring it does not exceed buffer size.
+ *
+ * @param dest Pointer to the destination buffer.
+ * @param data_size Size of the destination buffer.
+ * @param fmt Format string.
+ * @param ... Additional arguments for the format string.
+ * @return Length of data that would be written, including null terminator,
+ *         or negative value on encoding error.
+ */
+inline static int snprintf_len(char *dest, size_t data_size, const char *fmt,
```
devYaoYH: This feels like it can easily go wrong, let's add some tests to catch this! Please add a `/src/utils/test` subdir and have a `snprintf_len.c` test script!

devYaoYH (approved):
LGTM!

PR #212, opened 2025-11-08
Python test to ensure ksy file matches becon_stats

microbites0:


lc0001coll:
Will look to reviving this and taking it on. Would help make sure our ksy file is good for use with TinyGS (we can ask the team to update the file if things are wrong, but testing is always good).

PR #214, opened 2025-11-13
Finally figured out the proper invocation of rom_reboot

devYaoYH:
A whole bunch of stuff that didn't work, looped back around to the one thing that did... offsetting by XIP_BASE to the new address just works :/


devYaoYH:
Note that the `pt.json` changes to add `link: ['a', 0]` was crucial to allowing the XIP_BASE + 0x42000 offset to work

PR #223, opened 2025-11-22
Added payload software to flight repository

QuackWifHat:


devYaoYH:
Let's just name the high-level folder in root repo as `payload` and add/migrate over a README.md describing this project subdirectory.

PR #235, opened 2026-01-24
Ground station pyscript refactoring

microbites0:
- Refactored ground station code.py into proper codebase and multiple files under gs folder
- Added command authentication and auth state tracking
- Added boot state tracking
- Added message ID incrementing
- Added unittesting

### DIFF
--- a/ground_station/cli.py
+++ b/ground_station/cli.py
@@ -1,0 +1,18 @@
+"""CLI entry point for pip-installed ground station.
+
+This wrapper exists because code.py shadows the stdlib 'code' module
+when installed as a package. CircuitPython devices use code.py directly.
+"""
+
+import os
+import sys
+
+# Ensure the ground station directory is on the path so that
+# code.py (and its sibling modules) are importable.
+sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
+
+from code import main  # noqa: E402
+
+
+def entry():
+    main()
--- a/ground_station/pyproject.toml
+++ b/ground_station/pyproject.toml
@@ -26,6 +26,7 @@ dependencies = [
     "adafruit-circuitpython-rfm9x",
     "adafruit-circuitpython-hashlib",
     "circuitpython-hmac",
+    "adafruit-blinka",
     "pydantic>=2.0.0",
 ]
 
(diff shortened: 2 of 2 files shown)


ground_station/.gitignore
```diff
@@ -0,0 +1,6 @@
+__pycache__/
+*.pyc
+gs_state.json
+.pytest_cache/
```
devYaoYH: Pytest may be shared between payload and ground station. If so, we can have this in the repo root .gitignore

ground_station/gs/comms.py
```diff
@@ -0,0 +1,169 @@
...
+                    print(f"  beacon_data_len={data_len}")
+                    
+                    # Extract beacon data payload
+                    if len(packet) >= 1 + data_len:
+                        beacon_payload = packet[1:1+data_len]
+                        beacon_data = protocol.decode_beacon_data(beacon_payload)
+                        beacon_data['raw_hex'] = packet.hex() if hasattr(packet, 'hex') else ''.join('{:02x}'.format(b) for b in packet)
+                        print(f"  State: {beacon_data['state_name']}")
+                        
+                        if beacon_data['stats']:
```
devYaoYH: This segment of the code looks repeated from `log_beacon` in logger.py. Consider whether it can be simplified by just calling the log function here?

ground_station/gs/hardware.py
devYaoYH: Consider renaming this to something more informative like radio_initialization.py

ground_station/gs/logger.py
```diff
@@ -0,0 +1,118 @@
+import os
```
devYaoYH: add some comments on the function of this logger? I think I might have mis-understood this as logging to console, but it actually writes into a .csv file instead?

One can also abstract the unpacking of beacon_data and allow for an argument into the logger function to say print to console, write to .csv, or both!

ground_station/gs/state.py
```diff
@@ -0,0 +1,62 @@
...
+        if beacon_boot_count != self.boot_count:
+            print(f"Syncing Boot Count: {self.boot_count} -> {beacon_boot_count}")
+            self.boot_count = beacon_boot_count
+            self.save()
+
+    def get_next_msg_id(self):
+        """Increment and return next message ID"""
+        self.msg_id += 1
+        # For a high-speed pass, we might not want to save on EVERY message
+        # but for now we'll keep it for safety unless performance is an issue
```
devYaoYH: Yeah, we might want to keep file IO to a minimum here. For context, file transfer protocol will require the ground_station to send on the order of tens of thousands of packets in a short duration of time so we cannot afford to open/close a file each time a new message gets sent.

For the purposes of logging, we may which to emit a message to some other queue consumption mechanism (think MQTT) to be processed offline (asynchronously). And here, we commit state once every 10 minutes or so for the ground station or upon shutdown etc. Once we receive a beacon or edit the saved state file with updated parameters, we should be able to formulate valid command packets (so correctness on startup is not too critical).

devYaoYH (commented):
Generally a good start to refactoring! Left a whole bunch of comments, do push back if anything sounds unclear, or more suitable as a follow-up!

PR #259, opened 2026-02-07
Bazel build system migration!

devYaoYH:
Migrate build system to use bazel with pick-sdk support via bzlmd

> [!CAUTION]
> This change is UNSAFE and should be tested on a PiCubed prior to pushing!

Lots of stuff has changed places and we need to ensure everything is linked up properly. This change will break the previous CMakeLists system!!

PR #264, opened 2026-02-20
made code compatible for pico

megargayu:
skipped over radio tasks and RBF pin detection. also added docs & build_pico.sh and fixed pico build for 23XX


megargayu:
We are going to go ahead with bazel migration as of #259, so we are not merging this right now.

Issue #268, opened 2026-02-21
[Testing/MRAM] Hardware test task mram driver is failing!!!


megargayu:
Currently, the driver we have for MRAM has never been tested. Since we have a hardware test task now, I've been able to run our tests for the driver (which have so far only been run on mocks) on actual hardware, and it's been failing! This is a huge issue, as if the driver doesn't work properly, then we can't use MRAM at all, so filesys and ftp are both out the window, as well as OTA and other interacting components.

The test can be found in `tasks/hardware_test/test/mram_test.c`

Two known issues so far:
1. `test_mram_write_disable_enable` - `mram_write_disable` doesn't seem to work, i.e. it allows writing even when disabled. Therefore, the test fails with:
```
[14:51:42.734] Waiting for tty device..
[14:55:29.090] Connected to /dev/ttyACM0
ERROR: Assertion failed: memcmp(buffer, "This should not be written", 26UL) != 0ERROR: Assertion failed: memcmp(buffer, "This should not be written", 26UL) != 0ERROR: Assertion failed: memcmp(buffer, "This should not be written", 26UL) != 0
```
2. `test_mram_preserve_data_on_sleep` - When commenting out the above test, this test simply never finishes:
```
[14:58:45.894] Waiting for tty device..
[15:00:04.091] Connected to /dev/ttyACM0
[DEBUG] main: Persistent data initialized, reboot count = 24
[DEBUG]       rebot_counter++ -> 24
[DEBUG] main: Slate uses 608 bytes
[INFO] main: Initializing...
[DEBUG] sched: Enumerated 4 tasks
[DEBUG] sched: Initializing task print
[INFO] Test task is initializing...
[DEBUG] sched: Initializing task blink
[DEBUG] sched: Initializing task hardware_test
[INFO] Hardware test task: running 1 test(s)
[INFO] === [1/1] mram_test ===
[DEBUG] Starting MRAM test

[INFO] [mram] Init start
[INFO] [mram] Init complete
[DEBUG] Testing MRAM write...

[DEBUG] Testing MRAM read...

[DEBUG] Read from MRAM: Hello, MRAM!

[DEBUG] Testing MRAM preserve data across sleep/wake...
```
With no output afterward (infinite loop?)

megargayu:
Closed by #335

PR #269, opened 2026-02-21
Testing/bazel hardware test task

megargayu:
added the hardware test task! however, the only provided test right now (`mram_test`) is failing, so this is being pushed to #268

### DIFF
--- a/bzl/hw_test_types.h
+++ b/bzl/hw_test_types.h
@@ -9,5 +9,5 @@ typedef void (*hw_test_fn)(void);
 typedef struct
 {
     const char *name;
-    hw_test_fn  run;
+    hw_test_fn run;
 } hw_test_entry_t;


megargayu:
The hardware test task has been completely changed based on the new design, #271  Docs have also been populated (with help from #274 so it still has some Cmake stuff in it, which needs to be fixed based on the aforementioned PR), but may be a bit cooked. Build has been verified on PICO with it running actual tests properly and normal host tests work properly as well.

src/filesys/test/mram_test.c
```diff
@@ -0,0 +1,179 @@
...
+void test_mram_check_collision(void)
+{
+    LOG_DEBUG("Testing mram_check_collision behavior...\n");
+    mram_allocation_init();
+
+    const uint32_t base = 0x000008;
+    const size_t len = 32;
+
+    /* Register a base allocation and ensure it succeeds */
+    ASSERT(mram_write(base, (const uint8_t *)"This should be written", len));
```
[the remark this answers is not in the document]
devYaoYH: The way that this is allocated on the stack could also be random jibberish (after the first 22char of the string), however, it does avoid a buffer overflow issue with the current implementation if mram_write reads for `len` number of bytes from the provided char ptr.

Please fix?

src/tasks/hardware_test/BUILD.bazel
```diff
@@ -0,0 +1,64 @@
+load("//bzl:defs.bzl", "hardware_integration_test_suite")
+
+package(default_visibility = ["//visibility:public"])
+
+# ── Shared assert header ──────────────────────────────────────────────────
+# Kept as a standalone target so *_hw_lib targets can depend on
+# hardware_test_assert.h WITHOUT creating a circular dep through
+# hardware_test_lib.
```
devYaoYH: This smells... Perhaps better suited for declaration within the `/src/test_infrastructure` directory instead?

src/tasks/hardware_test/BUILD.bazel
```diff
@@ -0,0 +1,64 @@
...
+    srcs = ["hardware_test_task.c"],
+    hdrs = ["hardware_test_task.h"],
+    includes = ["."],
+    deps = [
+        "//src/common",
+        "//src/slate",
+        "//src/scheduler:state_machine",
+        ":hardware_test_lib",
+    ] + select({
+        "//bzl:test_mode": [
```
devYaoYH: how is this target used when in `--config=tests` test_mode?

To my understanding this would be a non-test target (link to actual drivers) so this code path is essentially dead code?

src/tasks/hardware_test/hardware_test_task.c
```diff
@@ -0,0 +1,36 @@
...
+        hw_tests[i].run();
+    }
+
+    LOG_INFO("Hardware tests complete.");
+}
+
+sched_task_t hardware_test_task = {
+    .name = "hardware_test",
+    // To make sure we see the test results, keep testing every 10 seconds.
+    .dispatch_period_ms = 10000,
```
devYaoYH: This means we would need to make sure our tests are all idempotent! Which is a good practice we should be doing anyway (test setup/teardown functions).

However, might the persistent data/hardware interfere with some of these tests running repeatedly?

bzl/gen_hw_tests_header.py
```diff
@@ -0,0 +1,61 @@
...
+        "",
+        "/* ── Test function signature ──────────────────────────────────────── */",
+        "",
+        "typedef void (*hw_test_fn)(void);",
+        "",
+        "typedef struct",
+        "{",
+        "    const char *name;",
+        "    hw_test_fn  run;",
+        "} hw_test_entry_t;",
```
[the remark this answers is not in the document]
megargayu: It allows it to be automatically generated, which makes it easier to add tests - do you think we should just do it manually in hardware_test_task?

src/tasks/hardware_test/hardware_test_task.c
megargayu: It can lead to issues, but preferably the tests should only touch their own things (not deal with anything outside what they define), clean up, and then stop. (But we can definitely run into issues with e.g. 99% of MRAM being used by another task, and causing this to failing a test.)

HOWEVER, I've realized that the filesys tests do the exact opposite! (They clear MRAM on every test). I will add an issue on this later.

@devYaoYH An idea would also be to add a separate config state, HARDWARE_TEST, that only contains this task of unit tests. That would ensure no other tasks interfere with the tests, and then its up to the user (us 😛) to care about potential side effects of stuff. (How would we do this? Would it be another config state? How would we allow for build on PICO and PICUBED for the same thing?)

(But, for example, if tests automatically create an expected state by e.g. clearing MRAM, and then edit the state later, then we don't really care about the end result, as all tests should just be expected to not know the current state of the hardware. For example, this means that filesys tests are actually okay, as no other tasks will ever run anyway, and tests are known to be written agnostic of hardware state). So actually, I think the idea of HARDWARE_TEST might be the best.

megargayu:
Tl;dr of review: For now, this is fine, but I will keep the above conversation open and we can work on refactoring to make it more fluid later. Also created #284 as part of this.

PR #270, opened 2026-02-22
Implement changes to the payload to allow for camera multiplexer compatability, along with programs to allow for easier testing of take_photo and take_vid

lc0001coll:
- Added a config.txt file containing the config that _seems_ to have led to fixing the camera multiplexer
- Made changes to camera_utils.py and commands.py to be compatible with the camera multiplexer
- Fixed issues with SSDV not containing the needed <stddef> library to build, and fixed incorrect path construction in ssdv_utils.py
- Created take_photo_cli.py and take_vid_cli.py, which allow for simpler and more effective testing for taking photos and videos


devYaoYH (approved):
LGTM, perhaps raise a separate issue then work on also making sure the PiCubed side of the payload command is working?

PR #274, opened 2026-02-22
WIP - changed some instances of cmake

megargayu:
There are a lot of things that are still reminiscent of cmake, which need to be changed to bazel - especially docs! I need help on this, as I don't really understand how a lot of things work (this is very very WIP/lots has not been changed yet, I gave up) - @devYaoYH can you help?

The goal would be to not have "cmake" listed anywhere in the repo. (search for cmake across the entire repository!)


megargayu:
As per #269:

We should also update [main/src/test_infrastructure](https://github.com/stanford-ssi/samwise-flight-software/tree/main/src/test_infrastructure) with information from #269.

PR #292, opened 2026-02-28
add binary size build action

### LINKED ISSUE #287: [Actions] Quick test to check the binary size after every build
We need to check the binary size every time we build using Github Actions, as we just had a really big scare 😦 of too much file size.

This should essentially grab the file size of the generated `.uf2` file in bytes for every build task, and then assert to make sure its not greater than `150KiB`.

megargayu:
closes #287


devYaoYH:
Screenshot 2026-03-01 114118

Very nice! Could we selectively enable this for only the PICUBED-FLIGHT version of the build script?

PR #295, opened 2026-03-02
Yiheng/radio pins refactor

devYaoYH:
Added fixes for making our flight code compatible with the new [pico-hat hardware platform](https://github.com/Rivercraft911/SoftwareDebugPicoHat)

This change adds:

1. a new board definition `/boards/samwise_pico.h` with pico-specific pin definitions
2. a new .bazelrc configuration `--config=pico-hat` which enables more drivers than PICO

Fixed in this PR:

- neopixel bgr -> rgb color ordering (minor but reads nicer now) & pin initialization on pico-hat platform
- radio command sending on ground_station needs to strip first 4 bytes (radiohead header) and manually encode them as function params to `adafruit_rfm9x.send()`
- removed `watchdog_enable` call from our watchdog driver. This unnecessarily starts the software watchdog circuit even for non Try-Before-You-Buy builds
- updated flash target offset to work with `ota_mvp` partition table (for testing)
- extended beacon to once every minute for now (sending ~128 bytes is estimated to take ~200ms)

### DIFF
--- a/boards/samwise_pico.h
+++ b/boards/samwise_pico.h
@@ -75,7 +75,7 @@
 
 // pico_cmake_set_default PICO_FLASH_SIZE_BYTES = (2 * 1024 * 1024)
 #ifndef PICO_FLASH_SIZE_BYTES
-#define PICO_FLASH_SIZE_BYTES (2 * 1024 * 1024)
+#define PICO_FLASH_SIZE_BYTES (4 * 1024 * 1024)
 #endif
 // Drive high to force power supply into PWM mode (lower ripple on 3V3 at light loads)
 #define PICO_SMPS_MODE_PIN 23


devYaoYH:
@claude_doc

.bazelrc
```diff
@@ -26,13 +26,25 @@ build --disk_cache=~/.cache/bazel
...
+# Raspberry Pi Pico + Radio Hat Profile (Pico with UHF radio attached)
+build:pico-hat --define=profile=pico-hat
+build:pico-hat --platforms=//platforms:arm_cortex_m33_platform
+build:pico-hat --define=PICO_BOARD=samwise_pico
+build:pico-hat --define=PICO_PLATFORM=rp2350
+build:pico-hat --copt=-DPICO=1
+build:pico-hat --copt=-DPICOHAT=1
+build:pico-hat --@pico-sdk//bazel/config:PICO_STDIO_USB=True
+build:pico-hat --aspects=@pico-sdk//tools:uf2_aspect.bzl%pico_uf2_aspect
+build:pico-hat --output_groups=+pico_uf2_files
```
megargayu: similar to above, should pico_bin_files be here?

src/init/init.c
```diff
@@ -65,18 +67,30 @@ static bool init_drivers(slate_t *slate)
     gpio_put(SAMWISE_RF_SCK_PIN, 0);
 #endif
 
-    // Initialize device drivers if on PICUBED
-#ifndef PICO
+    // Initialize radio if on PICUBED or PICO with radio hat
```
megargayu: we should fix comments here to say PICO instead of PICUBED (picubed is not checked here!)

src/drivers/rfm9x/rfm9x.c
```diff
@@ -710,7 +710,6 @@ void rfm9x_init(rfm9x_t *r)
     gpio_set_function(r->spi_clk_pin, GPIO_FUNC_SPI);
     gpio_set_function(r->spi_tx_pin, GPIO_FUNC_SPI);
     gpio_set_function(r->spi_rx_pin, GPIO_FUNC_SPI);
-    gpio_set_function(17, GPIO_FUNC_SPI); // ???
```
megargayu: just to confirm, have these changes been checked across not only PICO Hat, but other hardware targets? Do we need to do so?

src/drivers/flash/flash.c
```diff
@@ -8,7 +8,7 @@
 
 #include "flash.h"
 
-#define FLASH_TARGET_OFFSET (0x00079000)
+#define FLASH_TARGET_OFFSET (0x00082000)
```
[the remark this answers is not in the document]
devYaoYH: Done! Thanks :)

PR #296, opened 2026-03-02
Added FSM state transition visualization

devYaoYH:
Extended the current style of running_state_test to the overall FSM transitions to be made for each build profile (flight, bringup, debug).


devYaoYH:
@claude_doc

devYaoYH:
@claude_doc

PR #303, opened 2026-03-06
Added error handling and handling failed commands in rpi main

### LINKED ISSUE #176: [Payload] RPi program crashes on exception, need to gracefully handle errors and retry or ignore failed commands
We should also check the behavior of the payload when the program actually crashes, does it reboot? Does rebooting it automatically start the payload software?

lc0001coll:
**GPIO init (line 24-27)** — catches and prints errors so a GPIO failure doesn't prevent the program from starting

**Logger setup (line 30-41)** — catches errors, falls back to boot_count = -1, and attempts a fallback setup_logger call

**Serial port (line 48-58)** — retries up to 5 times with 2s delay before giving up

**Main loop (line 67-71)** — the critical fix: wraps receive_and_dispatch_command() in try/except so any serial error or unexpected exception gets logged and the loop continues accepting commands


payload/main.py
```diff
@@ -15,28 +17,54 @@
...
+boot_count = -1
+try:
+    update_boot_count()
+    boot_count = get_boot_count()
+    setup_logger(boot_count)
+    clean_logs()
+except Exception as e:
+    print(f"Logger setup failed: {e}")
+    try:
+        setup_logger(boot_count)
```
devYaoYH: so this retries setup_logger up to 2 times? Could we make it cleaner with a set number of retires?

When would this fail?

Alternatively we can skip this as it doesn't appear too critical for the proper functioning of the payload's core logic? Only for when we're debugging?

devYaoYH (changes requested):
Thanks for improving the robustness of our payload! Couple comments for your follow-up

payload/main.py
lc0001coll: setup_logger might fail if the log directory isn't there, but that's about the only reason it might fail. I went with removing it since it's not super important to have.

megargayu (approved):
small changes/questions, but apart from those, LGTM!

payload/main.py
```diff
@@ -15,28 +17,52 @@
...
+boot_count = -1
+try:
+    update_boot_count()
+    boot_count = get_boot_count()
+    setup_logger(boot_count)
+    clean_logs()
+except Exception as e:
+    print(f"Logger setup failed: {e}")
 
 log = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
lc0001coll: logging.getLogger should be independent of setup_logger, so log should still be defined whether or not setup_logger fails

PR #305, opened 2026-03-06
Added 'beat cal!' into our beacon debug string

devYaoYH:
Added CALL TO ACTION!! debug string into our beacon

Tests to verify:

1. [flight code test] `bazel test //src/tasks/beacon:beacon_test --test_output=all` will print out the raw beacon hex bytes
2. copy over the hex dump bytes into `/src/ground_station/test_protocol.py`, something like:

```
6d 6f 63 6b 5f 73 74 61 74 65
20 62 65 61 74 20 63 61 6c 21
00 2a 00 00 00 39 30 00 00 00
00 00 00 00 00 00 00 00 00 00
00 00 00 00 00 00 00 00 00 00
00 00 00 00 00 00 00 a0 0f 00
00 00 00 00 00 00 00 00 00 00
00 00 00 00 00 00 80 3f cd cc
cc 3d cd cc 4c 3e 9a 99 99 3e
cd cc cc 3e 41 2a 00 00 00 4b
43 33 57 4e 59 00
```

3. [ground station test] `cd ground_station && uv run pytest tests/test_protocol.py` will assert that the hex dump can be decoded!


megargayu (approved):
LGTM! Beat cal!!!

PR #308, opened 2026-03-08
added the first malloc for filesys, added clear_and_init_slate()

### LINKED ISSUE #290: [MRAM] The first malloc...
We must now use the dreaded malloc(), for we are scared of multiprocessing.

I.e. there are two options for using the entirety of the RAM (or a big portion) for MRAM buffer:
1. Use Core 1 instead of Core 0, which allows to grow stack into heap directly
    * This would require threading, as we cannot directly boot into Core 1
2. Use malloc() and write to heap.

Therefore, we use #2, as we are more scared of threading than malloc. Essentially, just put a pointer to the array in the slate (with a lot of warning comments of course!), and malloc() on initialization of filesys. We will ONLY HAVE ONE MALLOC at the top of initialize! Do NOT realloc. Note that we do not need a free() as there is no destruction [we will just terminate the entire program!].

megargayu:
closes #290 (scary!)

Note that this hasn't been tested on hardware yet as it is blocked by #268 😦


src/filesys/filesys.c
```diff
@@ -78,6 +78,14 @@ filesys_error_t filesys_initialize(slate_t *slate, lfs_ssize_t *lfs_error_code)
         return FILESYS_ERR_MOUNT;
     }
 
+    free(slate->filesys_buffer);
```
devYaoYH: This looked scary, but `free(0)` is valid and a no-op, shall we leave a comment?

devYaoYH (approved):
LGTM! The one and ONLY!

devYaoYH:
Just revisiting this! Was playing around today with gemini cli while reviewing:

```
✦ On the RP2350, your memory is mapped across those three segments as follows:

  1. Where does BSS go?
  The BSS section goes into SRAM. 

  At boot, the RP2350's "runtime" (crt0) zero-initializes the region of SRAM designated for BSS. It does not take up any space in your Flash
  storage, but it occupies "live" memory once the program starts running.

  2. Overall Utilization Breakdown

  Based on your size output (text: 56300, data: 0, bss: 6068) and your hardware specs:


  ┌──────────────┬────────────────────┬───────────────┬──────────┬────────┐
  │ Segment      │ Components         │ Usage (Bytes) │ Capacity │ % Used │
  ├──────────────┼────────────────────┼───────────────┼──────────┼────────┤
  │ Flash (QSPI) │ text + data        │ 56,300        │ 500,000  │ 11.3%  │
  │ SRAM         │ data + bss + stack │ 10,164        │ 524,288  │ 1.9%   │
  └──────────────┴────────────────────┴───────────────┴──────────┴────────┘


  Flash Utilization (11.3%)
   * text: Your code and constants are stored here.
   * data: Usually stored in Flash and copied to RAM at boot. Since yours is 0, you are only storing instructions.
   * Remaining: You have ~443KB left for future code or filesystem storage (LittleFS).

  SRAM Utilization (1.9%)
   * bss: Your global variables (including the 1.6KB slate) are here (6,068 bytes).
   * stack: You mentioned a 4KB limit. The stack typically grows downward from the top of SRAM. At 4,096 bytes, your stack is roughly 2/3 the size
     of your current BSS.
   * data: 0 bytes.
   * Remaining: You have over 500KB of SRAM free. This is a massive amount of "headroom"—you could significantly increase your slate size or add
     large buffers (e.g., for image processing or radio packets) without worry.

  3. What happens when you "load" the binary?
   1. Flash: The .bin file (52KB) is written to the QSPI Flash.
   2. Boot: The RP2350's bootloader reads the Flash and starts executing.
   3. Runtime Init: The program immediately "carves out" space in SRAM for your BSS (6,068 bytes) and zeros it. It also reserves the Stack area
      (4,096 bytes).
   4. Execution: The CPU runs code directly from the Flash (XIP - Execute In Place), while reading and writing your slate variables in the SRAM.

  The Verdict: You are in a very healthy state. Your Flash is about 10% full, and your RAM is barely being touched (<2%). You have plenty of room
  to grow.

  1. Can BSS overshoot the Stack?
  At Link Time: No. The linker script for the RP2350 defines the SRAM layout. It assigns BSS to a fixed region and reserves a separate area for the
  Stack (usually at the very top of SRAM). If your BSS + Data is too large for the allocated SRAM, the build will fail with a "section .bss will
  not fit in region RAM" error.

  At Runtime: The risk isn't BSS "overshooting" the stack, but the Stack overshooting its own 4KB boundary. 
   * BSS is static; its size never changes.
   * Stack is dynamic; it grows downward as you call functions. 
  If you have deep recursion or large local variables (e.g., uint8_t buf[2048] inside a function), the stack will grow past its 4KB limit and start
  overwriting whatever is below it (which is usually the Heap or BSS). This is a classic "Stack Overflow" and causes a crash.
```

Some helpful commands to use:

Symbol sizes:
```
> nm --size-sort -S samwise.elf
```

Sums up pretty much into our .bin file size:
```
> nm --size-sort -S samwise.elf | perl -lane '$sum += hex($F[1]); END { print "Total: $sum bytes" }'
```

Stuff we put into .bss section of our RAM:
```
> nm --size-sort -S samwise.elf | grep " [Bb] "

20001de8 00000670 B slate
```

As we can see, the `slate_t` we declare is part of the `.bss` section of our compiled elf file and this region of memory gets 'reserved' in our RAM when the program loads, **it doesn't go into the stack!**

TL;DR we should revert the malloc call and just declare however much memory we need for the filesys buffer within the slate_t itself! @megargayu

Issue #87, opened 2025-05-25
[FSM] Add safe mode state for when battery power is low to lengthen mission time.


devYaoYH:
When certain criteria have been met (TBD - but at least low battery voltage/no charging current detected on fixed solar panels), we should enter some safe-mode which modifies the following behavior of our satellite:

- Increase interval between radio beacon sending
- Shut down ADCS + Payload RPi
- (Open Question) only enable MANUAL_OVERRIDE command?

devYaoYH:
This may depend on a minimum 50%? level of link error rate. I.e. if we do not have enough power, we may not want the radio to run at max power.

PR #313, opened 2026-03-13
Add a shutdown command & state to the picubed

### LINKED ISSUE #88: [Command] Add SHUTDOWN command.
We are legally required to add a shutdown command which should permanently disable communications with the satellite.

To prevent mistakes in triggering this, we should store a counter and only execute shutdown sequence if 3 consecutive SHUTDOWN commands are received.

lc0001coll:
Added a new shutdown command that, when received 3 times consecutively, puts the satellite into a shutdown state, where it doesn't respond to communications or transmit any signals. After 3 months in shutdown, the satellite will automatically exit the shutdown state and go back to the regular running state.


src/states/shutdown/shutdown_state.c
```diff
@@ -0,0 +1,29 @@
...
+state_id_t shutdown_get_next_state(slate_t *slate)
+{
+    neopixel_set_color_rgb(SHUTDOWN_STATE_COLOR);
+
+    if (slate->time_in_current_state_ms >= SHUTDOWN_TIMEOUT_MS)
+    {
+        LOG_INFO("Shutdown timeout reached (3 months). Reinitializing.");
+        slate->shutdown_triggered = false;
+        slate->shutdown_cmd_counter = 0;
+        return STATE_INIT;
```
megargayu: On FLIGHT, STATE_INIT will do burn wire. Presumably if we were able to get to shutdown state, wouldn't we have already done initialization & burn wire? Should we be moving to e.g. STATE_RUNNING?

src/states/running/running_state.c
```diff
@@ -3,6 +3,10 @@
 
 state_id_t running_get_next_state(slate_t *slate)
 {
+    if (slate->shutdown_triggered)
```
megargayu, line 6: If we shut down, wait 3 months, and then it goes back to running (or init -> burn wire -> running), then wouldn't this go back? 
a) Is that desirable behavior (infinite shutdown loop)? And if so:
b) Why did we pick 3 months before it just loops back? Is it just an arbitrary number?

megargayu (changes requested):
LGTM except for the two questions below--lmk if they make sense.

Also I thought shutdown meant like a completely system shutdown (i,e. no code running)? I think I am misunderstanding what shutdown means...

src/states/shutdown/shutdown_state.c
lc0001coll: Good point, will look into this, but I think you're right.

src/states/running/running_state.c
lc0001coll: The shutdown_triggered flag is set to false before the shutdown state returns the state to go back to after 3 months, so after 3 months, it'll run like normal again. In terms of why it's a 3 month timer, this was mostly out of suggestion from Yiheng when I was working on it. I'll go ahead and ping him if there's a specific reason for this timing that I might be forgetting.

megargayu:
To update this PR based on talks:

- "Shutdown" legally just means that we need to have total communication blackout (no in or out packets). This state must be persistent over reboots.
- We *can* reactivate with ground authorization, but otherwise, there should be NO WAY to reactivate.
(Specifically, FCC 47 CFR §97.113(a)(4)).

Therefore, this PR will work as it is. However, two things I think should be implemented:
1. We should add a way to get out of the shutdown state. This means checking for packets once in a while and only doing something on a specific one - otherwise, MAKE SURE to not send anything or perform any action!
2. I think it will be beneficial to investigate if we can move into a "low power state" or have a more efficient shutdown, as our current way of doing things will require a lot of unnecessary power draw/wear and tear on systems. Maybe if there is timer functionality on this low power state, move it to low power only for a bit before coming back on to check telemetry?

Note: This would also help solve #87.

src/states/running/running_state.c
megargayu: Update: 3 months doesn't matter, anything will work. We need to be able to get back into the codebase, see comment below.

megargayu (commented):
LGTM!! Thanks :)

megargayu (approved):

megargayu:
@claude Can you review this PR?

PR #324, opened 2026-03-19
added queue mock

megargayu:
We need a queue mock for tests, specifically related to FTP. This is a simple queue to stand in for pico-sdk's version. Note that pico-sdk's version is thread-safe, but we do not have that here.


devYaoYH (approved):
Looks good, let's just clarify pico-sdk's naming in the struct. (element_size, count, level)

Issue #254, opened 2026-02-03
[FTP/Testing] Validate that unit tests generated by AI are actually okay/not cooked


megargayu:
This issue is in relation to [ftp-continued](https://github.com/stanford-ssi/samwise-flight-software/tree/ftp-continued). Read the [FTP docs (in-progress)](https://github.com/stanford-ssi/samwise-flight-software/blob/ftp-docs/src/tasks/ftp/README.md) for more information. If modifying FTP code, create a new branch and PR to ftp-continued, not filesys-and-ftp or main.

With the rise of generative AI, many people like myself have started using it to write unit tests. This is not necessarily good. 

Anyway, the point is we have to look through the FTP unit tests and make sure they make sense. I can already tell there are going to be a lot of issues!

PR #326, opened 2026-03-21
First pass of FTP implementation!

### LINKED ISSUE #251: [FTP] Use new filesys stubs that propagate LFS error codes, and send packets based on new docs
This issue is in relation to [ftp-continued](https://github.com/stanford-ssi/samwise-flight-software/tree/ftp-continued). Read the [FTP docs (in-progress)](https://github.com/stanford-ssi/samwise-flight-software/blob/ftp-docs/src/tasks/ftp/README.md) for more information. If modifying FTP code, create a new branch and PR to ftp-continued, not filesys-and-ftp or main.

I've updated a lot of the filesys implementation (see https://github.com/stanford-ssi/samwise-flight-software/tree/ftp-and-filesys) to allow for LFS propogation. Someone needs to merge the newest version of `ftp-and-filesys` into `ftp-continued` and then fix the usage of Filesys in FTP so it works properly. 

Note that This includes changing the packet format to report LFS_ERROR_CODEs whenever they are described in the docs linked above!

megargayu:
Huge PR to implement FTP so far. This will close #251 and #254, and provide our actual implementation for FTP, finally!

Note there are still commands to implement and design, as listed by issues with label `feat:ftp`. Also, ground station needs to be implemented.


src/utils/buffer_utils.h
devYaoYH, line 1: This is pretty neat! I like it :)

src/tasks/ftp/test/ftp_test.c
```diff
@@ -0,0 +1,1605 @@
...
+    // Init success sends no packet
+    TEST_ASSERT(queue_is_empty(&slate->tx_queue),
+                "No packet should be enqueued on successful init");
+
+    return 0;
+}
+
+int ftp_test_real_file(slate_t *slate)
+{
+    const char *file_path = getenv("FTP_TEST_REAL_FILE_PATH");
```
devYaoYH, line 1369: We usually resolve runtime dependencies using a `runfiles` provider in bazel and inject runtime deps with the "data:" attribute.

So the ftp_test's bazel buildrule will look something like:

```
samwise_test(
    name = "ftp_test",
    srcs = [
        "ftp_test.c",
        "ftp_test.h",
    ],
    data = [
        ":generate_samwise_bin", # <-- `outs` the samwise .bin file
    ],
    deps = _FTP_DEPS,
)
```

Then in the code use runfiles resolution to read back the correct filepath to the generated output.

https://fuchsia.googlesource.com/fuchsia/+/HEAD/build/bazel/BAZEL_RUNFILES.md?format%2F%2F

HOWEVER! this point appears moot as bazel does not yet have support in C for this runtime generated file resolution lib :(

BUT, we can try leveraging `rlocationpath` which resolves the generated runtime filepath (as suggested by gemini):

```
# 1. The rule that generates the .bin file
genrule(
    name = "generate_data",
    outs = ["data.bin"],
    cmd = "echo 'binary data' > $@",
)

# 2. Your C binary depending on it
cc_binary(
    name = "my_app",
    srcs = ["main.c"],
    data = [":generate_data"], # Depend on the rule/output
    copts = ["-DBIN_PATH=\\\"$(rlocationpath :generate_data)\\\""],
)
```

So in this case, we avoid the need for the external run_tests wrapper and directly depend on the .bin buildrule from the test

src/drivers/rfm9x/BUILD.bazel
```diff
@@ -6,16 +6,16 @@ cc_library(
     srcs = ["rfm9x.c"],
     hdrs = ["rfm9x.h"],
```
devYaoYH, line 7: Perhaps a potential migration task in the near future (asked gemini cli about this):

  2. Consolidating _hdrs into cc_library
  You're correct that the current structure involves some redundancy in the driver BUILD.bazel files. Having the driver targets depend on their own _hdrs target is a valid refactoring that would reduce duplication:

```
    1 # Refactored example for src/drivers/rfm9x/BUILD.bazel
    2 cc_library(
    3     name = "rfm9x",
    4     srcs = ["rfm9x.c"],
    5     deps = [
    6         ":rfm9x_hdrs",  # Consolidate header and shared deps here
    7         "//src/utils",  # Implementation-only deps
    8     ],
    9     target_compatible_with = ["//platforms:arm_cortex_m33"],
   10 )
```

  This hasn't been done yet, likely because the current explicit structure was simpler to implement initially, but consolidating them as you suggested would be cleaner.

src/filesys/test/filesys_test.c
```diff
@@ -2359,157 +2342,79 @@ int filesys_test_read_multi_chunk_file_success(slate_t *slate)
...
+    {32, filesys_test_read_data_chunked_success, "Read Data - Chunked Read",
+     true},
+    {33, filesys_test_read_data_past_eof_success, "Read Data - Past EOF", true},
+    {34, filesys_test_read_file_seek_success, "Read File Seek", true},
+    {35, filesys_test_read_file_tell_success, "Read File Tell", true},
+    {36, filesys_test_read_file_size_success, "Read File Size", true},
+    {37, filesys_test_close_file_read_success, "Close File Read", true},
+    {38, filesys_test_read_full_workflow_success, "Read Full Workflow", true},
+    {39, filesys_test_read_multi_chunk_file_success, "Read Multi-Chunk File",
+     true},
```
devYaoYH: We removed these `true` in PR #319 eh? What's the order of dependent PR chains here? Shall we do 319 prior to this 326?

src/test_mocks/pico/util/queue.h
```diff
@@ -2,35 +2,80 @@
...
-// Mock queue functions that are no-ops in tests
 static inline void queue_init(queue_t *q, unsigned int element_size,
                               unsigned int element_count)
 {
+    q->element_size = element_size;
+    q->element_count = element_count;
+    q->head = 0;
+    q->tail = 0;
+    q->level = 0;
+    ASSERT(element_size > 0 && element_count > 0 &&
```
devYaoYH, line 21: Why do we use `TEST_ASSERT` in some places and `ASSERT` in others?

IIRC `ASSERT` is blocking and will result in the test runner script exiting?

.gitignore
```diff
@@ -72,10 +72,8 @@ cmake_install.cmake
 Makefile
 *.cmake
 
-# Build scripts (should not be ignored)
-!build_debug.sh
-!build_tests.sh
-!CMakeLists.txt
+# Run scripts (should not be ignored)
+!run_tests.sh
```
devYaoYH: Slightly surprised this gets ignored otherwise?

src/tasks/ftp/test/ftp_test.h
```diff
@@ -0,0 +1,65 @@
...
+int ftp_test_e2e_single_cycle_file(slate_t *slate);
+int ftp_test_e2e_multi_cycle_file(slate_t *slate);
+int ftp_test_e2e_cancel_then_new_file(slate_t *slate);
+int ftp_test_e2e_crc_mismatch(slate_t *slate);
+
+/* G. Init Tests (9X) */
+int ftp_test_init_success(slate_t *slate);
+int ftp_test_init_clears_state(slate_t *slate);
+
+/* H. Real File Test (10X) */
```
devYaoYH: nit: curious what does the '10X' mean here in these comments?

src/tasks/command/command_parser.h
```diff
@@ -21,13 +22,18 @@ typedef enum
...
     PAYLOAD_TURN_OFF,
-    MANUAL_STATE_OVERRIDE
+    MANUAL_STATE_OVERRIDE,
+
+    /* FTP Commands */
+    FTP_REFORMAT,
+    FTP_START_FILE_WRITE,
+    FTP_WRITE_TO_FILE,
+    FTP_CANCEL_FILE_WRITE,
+    FTP_FORMAT_FILESYSTEM,
```
devYaoYH: Where are we describing the functionality of these commands?

src/common/config.h
```diff
@@ -14,8 +14,25 @@
 /**
  * FTP Configuration
  */
+// Time to wait between dispatches when idle
```
devYaoYH, line 17: What is a 'dispatch'? When the satellite replies which messages have been received during an active pass over the N number of packets?

devYaoYH (commented):
brief initial comments first, still reviewing the core FTP_TASK against design doc specs!

src/tasks/ftp/test/ftp_test.h
megargayu: It means the ID numbers for the tests - for example, all initializing tests have codes `9X` (90 and 91), while all write file data tests have codes `3X` (30, 31, ...).

src/tasks/command/command_parser.h
megargayu: FTP docs - `ftp/README.md`.

src/common/config.h
megargayu: A `dispatch` just means the dispatch of the FTP task. I coded it so that FTP runs slower whenever it is not uploading a file, and runs faster after `start_file_write` is run, to save on resources.

A better way might also be to "wake up" FTP on start file write from `command_parser` rather than checking in a queue using a task - but this is good enough for now.

I know the idea of an FTP state was also floated around, which might be worth investigating when we get more tasks in (as right now I don't even think that much should be disabled in an FTP state).

src/test_mocks/pico/util/queue.h
megargayu: This `ASSERT` should never run (like never run, if it is run then there is something really wrong with the mock), and the reason why its used is because `TEST_ASSERT` is made mainly for test files where there is a return value (e.g. `-1`). If this queue does fail then we do need to panic as basically everything will break, and we can't change the signature of the function to support error code passthrough as we need to keep the same API as `pico-sdk`.

src/tasks/ftp/test/ftp_test.c
megargayu: I've used claude to generate this :) It did not use a genrule, but instead used a "transition" under `//bzl/firmware_for_test.bzl`. This also meant it had to move our `define`s in the build config to `flag`s, but this should have no impact on the code (I think).

```md
Recommendation                                                                                                                                                                                                                                                 

  Use a Starlark rule with an outgoing configuration transition that builds :samwise under the picubed-debug settings and re-exports the resulting .bin as a host-config file. Wire that target into the FTP test's data attribute and use $(rootpath …)
  expansion in the test's env attribute. After this, plain bazel test //... works and run_tests.sh becomes redundant.

  Why a transition (not a genrule)

  - :samwise has target_compatible_with = ["//platforms:arm_cortex_m33"], so it can't be a plain data dep of a host cc_test.
  - A genrule builds srcs in the target config and tools in the exec config — neither gives us arm_cortex_m33_platform + the rest of --config=picubed-debug (profile=picubed-debug, PICO_BOARD, PICO_PLATFORM, compilation_mode=dbg, -DDEBUG=1, -Og,
  PICO_STDIO_USB=True). A transition is the only Bazel-native way to switch into that configuration mid-graph.
  - The raw .bin is produced by //bzl:bin_aspect.bzl%pico_bin_aspect, which today only runs because --aspects=... is in .bazelrc for that config. We have to attach that aspect ourselves through the transition rule's label attribute.
  ```

File: src/tasks/ftp/test/ftp_test.h (hunk 6 above)
Comment (devYaoYH) [reply src/tasks/ftp/test/ftp_test.h]: Where are these test codes used? We should clarify this in the comments.

File: src/tasks/command/command_parser.h (hunk 7 above)
Comment (devYaoYH) [reply src/tasks/command/command_parser.h]: We weren't previously disciplined, but imo we should at least write a short comment inline with a link to the definitions?

File: src/common/config.h (hunk 8 above)
Comment (devYaoYH) [reply src/common/config.h:17]: Why do we need to run the FTP task in the running_state? Should we not move into another ftp_state upon receiving a file transmission request?

This dynamic adjusting of scheduling makes it harder to reason about the FSM transitions... Even if ftp_state has a large overlap with running_state, it is at least easier to map out the transitions our satellite makes in order to receive a file. What is the advantage of keeping ftp_task within running_state?

File: src/tasks/ftp/test/ftp_test.c (hunk 1 above)
Comment (devYaoYH) [reply src/tasks/ftp/test/ftp_test.c:1369]: The compatibility with target_platform is a good catch by claude. The final solution feels somewhat convoluted and clunky... but without additional layering on top of bazel, perhaps the best we can get that directly chains these two build steps together... 🤔

File: src/common/config.h (hunk 8 above)
Comment (megargayu) [reply src/common/config.h:17]: The idea was basically that FTP is a really slow process, and it subject to multiple orbits to complete a file write. In the intermediary, we need to be able to run all other tasks in running_state, like beacon and radio and whatnot. 
1. I don't know what would trigger a transition between `running_state` and a hypothetical `ftp_state` - of course start file write will transition into `ftp_state` and cancel/complete would transition back, but would it come back if there are no packets after a long time? (We can't just wait 20 minutes in FTP state doing nothing).
2. What would be the differences in `ftp_state` and `running_state`? Almost all tasks need to be run alongside FTP, simply because it takes so long. This especially includes radio and command, but also beacon, watchdog, and telemetry - which turns out to be basically all tasks in running state.

Let me know if this makes sense. Theoretically if we only use FTP for OTA then we can reason that ftp_state could exist as a "degraded" state that solely exists for moving into OTA and causing a reboot into the new binary, but I think it's better to just keep FTP generalized for anything as it takes so long. (We also plan to use FTP for Payload in the future so we have one unified framework for sending/receiving files).

I can also see setting "times" when the FTP state powers on during a file write and allows for extreme focus on FTP, and then transitions back to `running_state` when the satellite is out of range. (Would this be possible or useful? I feel `running_state` is already pretty lightweight to not interfere with FTP's dispatch).

Do you want me to remove the dynamic dispatching? What you're saying makes sense for making sense of FSM transitions, but I also think it's useful to reduce operations (I guess the FTP task will literally just exit every time it comes in after checking queue, so it's very minimal amount of operations wasted, but it runs so quickly to capture file packets that we can still consider it).

File: src/tasks/command/command_parser.h (hunk 7 above)
Comment (megargayu) [reply src/tasks/command/command_parser.h]: done!

File: src/tasks/ftp/test/ftp_test.h (hunk 6 above)
Comment (megargayu) [reply src/tasks/ftp/test/ftp_test.h]: done!

**Comment by megargayu** [comment]
Update: I was able to test this on the `mram` with the new mram drivers cooked up by @iceiscool456, and it worked!!! So FTP officially works on all unit tests on real hardware (yippee)!

File: src/tasks/ftp/ftp_task.c [hunk 9]
```diff
@@ -0,0 +1,590 @@
...
+    uint8_t status_report_data[PACKET_DATA_SIZE];
+    size_t len = buffer_encode_data(
+        status_report_data, BUFFER_ENCODE_ARG(slate->ftp_start_cycle_packet_id),
+        BUFFER_ENCODE_ARG(ftp_get_last_packet(slate)),
+        BUFFER_ENCODE_RAW(FTP_PACKET_TRACKER_SIZE,
+                          slate->ftp_packets_received_tracker),
+        BUFFER_ENCODE_ARG(file_crc_so_far),
+        BUFFER_ENCODE_ARG(total_bytes_written),
+        BUFFER_ENCODE_ARG((slate->filesys_buffer != NULL) ? 1 : 0),
+        BUFFER_ENCODE_ARG(slate->filesys_is_writing_file ? 1 : 0));
```
lc0001coll: In the README on lines 423 and 424, filesys_buffer_malloced and filesys_is_writing_file are listed as uint8_t, but this looks like they'd be regular ints unless they're being casted somewhere else.

src/tasks/command/command_task.c
```diff
@@ -23,6 +25,15 @@ void command_task_init(slate_t *slate)
...
 
+    queue_init(&slate->ftp_start_file_write_data,
+               sizeof(FTP_START_FILE_WRITE_DATA), FTP_DATA_CAPACITY);
+
+    queue_init(&slate->ftp_write_to_file_data, sizeof(FTP_WRITE_TO_FILE_DATA),
+               FTP_DATA_CAPACITY);
+
+    queue_init(&slate->ftp_cancel_file_write_data,
+               sizeof(FTP_CANCEL_FILE_WRITE_DATA), FTP_DATA_CAPACITY);
+
```
lc0001coll, line 39: I think ftp_format_filesystem_data should also be added here since we use queue_try_add on it in the ftp task.

src/tasks/ftp/ftp_task.h
```diff
@@ -0,0 +1,356 @@
...
+     *
+     *     Received (FTP_PACKET_TRACKER_T) - Bitfield of
+     * received packets in this set.
+     *
+     *     File_CRC_So_Far (FILESYS_BUFFERED_FILE_CRC_T) - CRC computed over
+     * bytes written to MRAM so far (0 if first cycle).
+     *
+     *     Total_Bytes_Written (FILESYS_BUFFERED_FILE_LEN_T) - Total bytes
+     * written to MRAM so far.
+     *
```
lc0001coll, line 165: filesys_buffer_malloced is missing here

src/tasks/ftp/ftp_task.c
```diff
@@ -0,0 +1,590 @@
...
+    {
+        LOG_ERROR("[FTP] Failed to write buffer to MRAM: %d", mram_write_res);
+        // Send error packet with the error code
+        send_ftp_lfs_error_packet(slate, FTP_FILE_WRITE_MRAM_ERROR,
+                                  mram_write_res, mram_lfs_err);
+        return;
+    }
+
+    // Reset received tracker for next cycle
+    ftp_tracker_clear(&slate->ftp_packets_received_tracker);
```
lc0001coll, line 368: If mram write fails and returns without clearing the tracker, could the duplicate check in the tracker stop us from retransmitting the failed files?

src/tasks/command/command_task.c
lc0001coll: Also, these queues would need to be freed in free_slate.

src/common/config.h
lc0001coll: Based on the design, I'd agree that ftp_state wouldn't be much more different than running_state, and I think the extra complexity of deciding how to handle transitioning from the ftp state to running state might not bring alot of benefit.

lc0001coll: Also, by what I can see ftp_task isn't in the running state in the current PR. Was it taken off?

bzl/firmware_for_test.bzl
```diff
@@ -0,0 +1,64 @@
+"""Build the picubed-debug firmware .bin from a host-config test.
+
+`firmware_bin_for_testing` wraps a cc_binary (typically `//:samwise`) and
+exposes its raw `.bin` image as a regular file usable as `data` on a host
+`cc_test`. A Starlark transition switches into the same configuration that
+`--config=picubed-debug` sets up in `.bazelrc`, and `pico_bin_aspect` (which
+runs under the transitioned config, so it has the arm-none-eabi toolchain)
+performs the ELF→bin objcopy.
+
+If you change `build:picubed-debug` in `.bazelrc` (lines 54-65), mirror the
```
lc0001coll, line 10: Just wondering, is there any kind of thing in bazel or another tool that could ensure the builds are the same or atleast test them? Would be an extra nice to have just to make sure it doesn't bite us later.

src/tasks/ftp/ftp_task.c
lc0001coll: nvm, looked at buffer_utils again and it looks like they'd be casted into uint8_t. Cool stuff!

src/README.md
```diff
@@ -32,7 +32,7 @@ bazel-bin/samwise.uf2
 ### Running tests
 
 ```bash
-bazel test //...
+source run_tests.sh
```
lc0001coll: nit: run_tests.sh was deleted but is listed here

src/common/config.h
megargayu: Good catch! Just fixed :)

src/tasks/ftp/ftp_task.c
megargayu: I'm not sure exactly what you mean. One thing I have realized (is this what you are asking) is that if we complete a cycle, then we error out and don't accept any more packets, but also cannot write to MRAM. We can add a new packet command that just tries to write to mram (essentially second half of the process_file_write_data_command, which will then be called by the original function) so we can write multiple times if an error occurs. Do you think this is the best way to go about it?

bzl/firmware_for_test.bzl
megargayu: What do you mean by 'ensure the builds are the same'? Do you mean a byte-by-byte comparison? We have unit tests and can test integration/other stuff on the ground before FTP/OTA.

lc0001coll: I shoulda been a bit more specific. Moreso was just referring to having some kind of automatic check that the relevant flags in this file are the same as in .bazelrc.

src/tasks/ftp/ftp_task.c
lc0001coll: In process_file_write_data_command, we have this duplicate packet check

```// Check for duplicate packet
    if (ftp_tracker_check_bit(&slate->ftp_packets_received_tracker,
                              packet_index))
    {
        LOG_INFO("[FTP] Duplicate packet ID %d received, ignoring.",
                 command_data.packet_id);
        return;
    }
```
My worry was that if we run into an error on writing to the mram and return without clearing the tracker, if we try writing again after the error, the already written bits will still be within the tracker and be flagged as duplicates. 

For now, this actually might not be an issue since the failed write deletes the file and causes the FTP_ERROR_NOT_WRITING_FILE error to pop up until we start a fresh file write which clears the tracker anyways. 

If we do end up having a new command that writes to mram after an error, we'd still get an error since we set ```filesys_is_writing_file = false``` and clear the buffer after a write failure, and filesys_write_buffer_to_mram first checks ```if (!slate->filesys_is_writing_file)``` and ```if (!slate->filesys_buffer_is_dirty)``` before doing anything. If we pass these checks, then the tracker duplicate check could become an issue.

PR #336, opened 2026-04-25
Adcs packet

nablaxcroissant:
Running pr for adcs <-> picubed.


ground_station/radio_initialization.py
```diff
@@ -33,11 +33,11 @@ def get_board_pins(board_type):
...
-            "MOSI": board.GP15,
-            "MISO": board.GP12,
-            "SCK": board.GP14,
-            "CS": board.GP13,
-            "RESET": board.GP11,
+            "MOSI": board.GP19,
+            "MISO": board.GP16,
+            "SCK": board.GP18,
+            "CS": board.GP17,
+            "RESET": board.GP21,
```
megargayu, line 40: Are these pins universal? @lc0001coll should check 👍

src/drivers/adcs/adcs_driver.h
```diff
@@ -22,33 +21,33 @@ typedef enum
  * @param slate Pointer to slate structure
  * @return ADCS_SUCCESS on success, error code otherwise
  */
-adcs_result_t adcs_driver_init(slate_t *slate);
```
megargayu, line 25: OK Carson I understand why you removed slate and wanted structs for each component instead. I think it may be better to do that, as we want to avoid static variables as much as possible. @nablaxcroissant Can we make these take in a pointer to a "state struct", which is then stored by code that interacts with adcs to the slate?

src/tasks/command/command_parser.h
```diff
@@ -21,7 +21,10 @@ typedef enum
     PAYLOAD_EXEC,
     PAYLOAD_TURN_ON,
     PAYLOAD_TURN_OFF,
-    MANUAL_STATE_OVERRIDE
+    MANUAL_STATE_OVERRIDE,
+    PAYLOAD_SHUTDOWN,
```
megargayu: Why is PAYLOAD_SHUTDOWN in this PR? It's not even added into `command_parser.c`? Lets remove this and focus on this in another PR (I know @lc0001coll is working on SHUTDOWN as well).

src/tasks/command/command_parser.h
```diff
@@ -21,7 +21,10 @@ typedef enum
     PAYLOAD_EXEC,
     PAYLOAD_TURN_ON,
     PAYLOAD_TURN_OFF,
-    MANUAL_STATE_OVERRIDE
+    MANUAL_STATE_OVERRIDE,
+    PAYLOAD_SHUTDOWN,
+    ADCS_EXEC,
+    ADCS_PACKET
```
megargayu, line 26: Same with this - if we can implement ADCS_EXEC and ADCS_PACKET properly in this PR before merging that will be much cleaner and easier to track :)

src/drivers/adcs/adcs_driver.h
megargayu: In fact it actually may be better to do this in another PR and refactor how we do slate stuff so everything is updated in this regard. I made a new issue to track this, #337.

ground_station/radio_initialization.py
lc0001coll: These pins work on the radio boards I've tested on 👍 For the breadboard-based ground stations, the wiring would just have to be moved around a little bit but it should work for them too.

src/tasks/adcs/adcs_task.c
```diff
@@ -11,48 +11,76 @@
...
 
-        if (slate->adcs_num_failed_checks >=
-            ADCS_MAX_FAILED_CHECKS_BEFORE_REBOOT)
+    LOG_INFO("[ADCS] TX COUNT {%d}", tx_count);
+    if (uart_comms_packet_ready(SAMWISE_ADCS_UART))
+    {
+        LOG_INFO("[ADCS] PACKET RECEIVED {%d}", rx_count);
+        rx_count += 1;
+        msg_t received;
+        receive_msg(&received, rx_buf);
```
devYaoYH: CMIIW, `receive_msg` will get called once each dispatch (1s)? The code path I'm seeing is:

receive_msg -> uart_comms_get_packet -> (reads from) rx_buffer
irq writes into rx_buffer (cyclic buffer, UART_RX_BUFFER_SIZE 256 for now)

which is okay if we keep below around 256 bytes within a second window.

Also what happens if malformed bytes arrive over the wire? Does the get_packet throw them away if it isn't parsed (e.g. increment the head ptr index)? Or we do some checking/filtering during ingestion time (don't do this in the irq tho).

nablaxcroissant: Hmm yeah it looks there is a bug where if you fill 256 bytes without a zero at will get stuck and never return packet ready. I think we would actually prefer to have buffer corruption than for that to happen. In the case that buffer corruption happens, it doesn't get parsed unless the buffer happens to hit onto one of the packet flags so if like the 5th byte happens to be 0-5. I think worst case behaviour it would read past the bounds of the received packet like for example it would read an adcs packet from garbage memory. This is fixable by just making sure that for fixed size packets the packet isn't shorter than it should be. So there should be 3 fixes

1. Fix the rx so that if 256 nonzero bytes enter it doesn't get stuck.
2. Add checksums. I already have a byte reserved for crc8 checksum, I just didn't implement it yet.
3. For fixed-size packets we just check packet size. For non-fixed size we can just tolerate corruption, we aren't actually using those yet and nor should we for anything important.

Most other cases will be fine. If it doesn't match a know packet type it will just get discarded. Everything should work right now if the sender is doing things normally but if a byte gets dropped it could do weird stuff.  The only thing here that would be fatal would be dropping the null termination byte and overflowing the buffer.

src/drivers/adcs/adcs_driver.c
```diff
@@ -73,61 +80,59 @@ static uint32_t adcs_driver_read_uart_with_timeout(char *buf,
...
-        SAMWISE_UART_RX_FROM_ADCS,
-        UART_FUNCSEL_NUM(SAMWISE_ADCS_UART, SAMWISE_UART_RX_FROM_ADCS));
-
-    // Set data format
-    uart_set_format(SAMWISE_ADCS_UART, ADCS_UART_DATA_BITS, ADCS_UART_STOP_BITS,
-                    ADCS_UART_PARITY);
+    flush_uart();
 
+    // Initialize uart for sending commands/receiving telemetry
+    // uart_init(SAMWISE_ADCS_UART, ADCS_UART_BAUD);
```
megargayu: Why was this completely commented out? Good to delete if not necessary anymore

src/drivers/adcs/adcs_driver.c
```diff
@@ -161,18 +166,23 @@ adcs_result_t adcs_driver_get_telemetry(slate_t *slate, adcs_packet_t *packet)
     }
 
     // Return true if we received all expected bytes
-    return slate->is_adcs_telem_valid ? ADCS_SUCCESS : ADCS_ERROR_UART_FAILED;
+    return is_adcs_telem_valid ? ADCS_SUCCESS : ADCS_ERROR_UART_FAILED;
```
megargayu, line 153: We can remove this ternary operator as we already return in the if (!is_adcs_telem_valid) branch

src/drivers/adcs/adcs_driver.c
```diff
@@ -27,6 +31,9 @@
 // currently set to 500ms (faily generous)
 #define ADCS_BYTE_TIMEOUT_US (500000)
 
+static bool is_adcs_on = false;
```
megargayu, line 34: We should consider whether we want to have static state. I think its much better if we basically take in a struct by reference for  every function (`adcs_state_t &adcs_state`) and then have this on the `slate_t`.

src/drivers/communications/uart_communications.h
```diff
@@ -0,0 +1,63 @@
...
+ * @return Number of bytes copied to buffer
+ */
+uint16_t uart_comms_get_packet(uart_inst_t *uart_instance, uint8_t *buffer,
+                               uint16_t max_length);
+
+/**
+ * UART interrupt handler callback
+ *
+ * @param uart_instance UART peripheral instance number
+ */
```
megargayu: we should remove uart_instance from this docstring

src/packet/adcs_packet.h
```diff
@@ -22,6 +22,14 @@ typedef struct __attribute__((packed))
     // Quaternion Estimate
     float q0, q1, q2, q3;
 
+    // Time
+    float mjd;
```
megargayu, line 26: Why are we storing MJD with a float if we also have UTC_time? should it be preferred to use a long or an int?

megargayu: Same with UTC_time. If we are already using 64 bits for the combination of these two values, we could use Unix Timestamps (i.e. C's `time_t`) or some other more compressed data format (like, we can do an offset 32-bit Unix Timestamp - but anyway, a 32-bit Unix Timestamp will end in 2038, so this is even an option). MJD is good because we don't need historical data, so we can similarly use that assumption to cut this down.

For reference, I am talking about this to help alleviate issues down the road with [#357](https://github.com/stanford-ssi/samwise-flight-software/issues/357). And floats can have precision issues, so its probably not a good idea to use it here.

src/states/init/init_state.c
```diff
@@ -11,13 +11,15 @@ state_id_t init_get_next_state(slate_t *slate)
...
     // Check if RBF pin is detected
-    if (is_rbf_pin_detected())
-    {
-        // If RBF pin is detected, block and stay in init state
-        LOG_INFO("RBF pin detected, staying in init state.");
-        neopixel_set_color_rgb(0xff, 0, 0);
-        return STATE_INIT;
-    }
+    // TODO: Uncomment for updated model
+    return STATE_RUNNING;
```
megargayu, line 15: Should this still be commented?

src/states/running/running_state.c
```diff
@@ -31,11 +31,11 @@ sched_state_t running_state = {
...
-                               .num_tasks = 6,
-                               .task_list = {&print_task, &watchdog_task,
-                                             &beacon_task, &telemetry_task,
-                                             &radio_task, &command_task},
-                               .get_next_state = &running_get_next_state};
+sched_state_t running_state = {
+    .name = "running",
+    .id = STATE_RUNNING,
+    .num_tasks = 6,
+    .task_list = {/*&print_task,*/ &watchdog_task, &beacon_task,
```
megargayu: Should this still be commented out?

src/tasks/adcs/adcs_task.c
```diff
@@ -7,52 +7,88 @@
...
 #define ADCS_MAX_FAILED_CHECKS_BEFORE_REBOOT (5)
 
 void adcs_task_init(slate_t *slate)
 {
-    adcs_driver_init(slate);
 
-    slate->adcs_num_failed_checks = 0;
+    uart_comms_init(SAMWISE_ADCS_UART, SAMWISE_UART_TX_TO_ADCS,
+                    SAMWISE_UART_RX_FROM_ADCS, 115200);
+    // adcs_driver_init();
```
megargayu, line 27: I don't think adcs_driver_init() is run anywhere anymore. Is this expected behavior?

src/tasks/adcs/adcs_task.c
```diff
@@ -7,52 +7,88 @@
...
 {
     neopixel_set_color_rgb(ADCS_TASK_COLOR);
 
-    // Check if the board is alive
-    if (!adcs_driver_is_alive(slate))
-    {
-        // Not alive - increment persistence counter and reboot if persistent
-        slate->adcs_num_failed_checks++;
+    // Turn on adcs_pin after init for some reason
+    // TODO: figure out why this breaks the code if it happens during init
```
megargayu, line 44: how important would this bug be? can u make an issue?

src/tasks/adcs/adcs_task.c
```diff
@@ -7,52 +7,88 @@
...
+                    break;
+                LOG_INFO("[ADCS] Ping received");
+                send_pong();
+                break;
+            case MSG_PONG:
+                if (num_bytes != 8)
+                    break;
+                LOG_INFO("[ADCS] Pong received");
+                // don't send ping else we get infinite loop
+                slate->is_adcs_on = true;
```
megargayu, line 69: `slate->is_adcs_on` is now completely discoupled from the `static is_adcs_on` in adcs_driver.c!!! Should we remove `is_adcs_on` from the slate? What is the use of this code?

src/tasks/adcs/adcs_task.c
```diff
@@ -7,52 +7,88 @@
...
+                break;
+            case MSG_ADCS_PACKET:
+                if (num_bytes != 7 + 41)
```
megargayu: Can we calculate this without using magic numbers, so it doesn't get cooked if we change the adcs packet in the future?

src/tasks/adcs/adcs_task.c
```diff
@@ -7,52 +7,88 @@
...
+                    ;
```
megargayu: lets add a log here or something in the if statement

src/tasks/command/command_parser.h
```diff
@@ -21,7 +21,10 @@ typedef enum
     PAYLOAD_EXEC,
     PAYLOAD_TURN_ON,
     PAYLOAD_TURN_OFF,
-    MANUAL_STATE_OVERRIDE
+    MANUAL_STATE_OVERRIDE,
+    PAYLOAD_SHUTDOWN,
+    ADCS_EXEC,
+    ADCS_PACKET
     // add more commands here as needed
```
megargayu, line 27: lets remove `PAYLOAD_SHUTDOWN` and `ADCS_PACKET` if unnecessary

src/packet/adcs_packet.h
megargayu: Huh, looking into it it looks like other libraries use two `double`s/`float`s, similar to this approach, which is also probably what adcs outputs. I think then this makes sense, so nevermind!

src/drivers/adcs/adcs_driver.c
megargayu: I forgot we moved this to #337

PR #339, opened 2026-05-02
Flight build pr fix

lc0001coll:
PRs changing the main README or .yml files don't get the build workflow called for them. However, the PICUBED-FLIGHT build is still a required check that blocks them from approval. This PR should patch this.


megargayu (approved):
LGTM tysm

PR #340, opened 2026-05-02
some refactoring of ground station, distinguishing RX packets 

### LINKED ISSUE #338: [Ground Station] No way to distinguish RX packets!
We need a byte to distinguish RX packets, meaning everything from SAMWISE --> Ground Station needs to be redesigned 😄

megargayu:
1. Started refactoring ground station so it isn't so messy (main branch to track is [`ground_station/v2`](https://github.com/stanford-ssi/samwise-flight-software/tree/ground_station/v2)).
2. Added "Pippin" mode that doesn't do anything right now, for S-Band - See [Ground Station Spec Sheet](https://docs.google.com/document/d/1cDkb7PT_642zS3IgYS5149CFJi5kesSO_I0K0GwfCvY/edit?tab=t.0).
3. Added 1 byte at the beginning of each packet from SAMWISE --> Ground Station to distinguish between packet types - right now we only decode beacons!


ground_station/ground_station/main.py
```diff
@@ -0,0 +1,29 @@
...
+    if mode == "server":
+        from ground_station.server import run
+
+        run()
+    elif mode == "code":
+        from ground_station.code import main as code_main
+
+        code_main()
+    else:
+        raise SystemExit(f"unknown mode: {mode!r} (expected server|cli|code)")
```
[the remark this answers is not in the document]
lc0001coll: Bumping this, the cli is said to be an option in the file header comment and error message but doesn't have an actual path in the if statement.

ground_station/ground_station/radio_initialization.py
```diff
@@ -56,7 +56,7 @@ def get_board_pins(board_type):
             "SCK": board.SCK,  # Usually D24
             "CS": board.CE1,  # Lifted from run_sequence.py
             "RESET": board.D25,  # Lifted from run_sequence.py
-            # 'LED': board.D13        # Built-in LED
+            "LED": board.D13,  # Built-in LED
```
lc0001coll: Will test this on the payload, need to see if anything else like the camera is using this pin and causes a conflict. Hopefully not since this would help fix line 98 that crashes from trying to set up the led that isn't defined for the rpi.

ground_station/ground_station/models/packet_model.py
```diff
@@ -0,0 +1,62 @@
+from . import USE_PYDANTIC, _BaseModel
```
lc0001coll: When using a circuitpython without pydantic, _baseModel isn't defined and imports like this one will fail. Is this intended?

ground_station/ground_station/radio_commands.py
```diff
@@ -0,0 +1,242 @@
...
+
+            # All received packets are from the satellite. The first byte is a
+            # DownlinkPacketId (mirror of the uplink cmd_id) — dispatch on it.
+            if len(packet) < 1:
+                logger.warning("Empty packet received")
+
+                print(">>> END PACKET <<<\n")
+                return None
+
+            packet_id = packet[0]
```
lc0001coll: Have similar concerns as what copilot is saying, as of now, radio_task encodes the first 5 bytes of the packet to be [dst][src][flags][seq][len], and [receive strips the first 4 bytes](https://docs.circuitpython.org/projects/rfm9x/en/latest/api.html#adafruit_rfm9x.RFM9x.receive), so it seems like packet[0] would get the length of the packet rather than the ID.

lc0001coll (changes requested):
Left a couple comments on some things I think may need changes or clarification on, but overall lots of amazing changes and a great help to the ground station!

ground_station/ground_station/models/packet_model.py
megargayu: This should be fine, as:
```
if not IS_CIRCUITPYTHON:
    try:
        from pydantic import BaseModel

        _BaseModel = BaseModel
```

ground_station/ground_station/radio_commands.py
megargayu: Yeah this was completely cooked. I fixed it and testing, the entire packet looks good now! I want to refactor all of this so its not so confusing, because right now the way we decode things is really inconsistent which makes it hard to make changes like this :frowning:

PR #355, opened 2026-05-20
fixed ground station after update!

megargayu:
finally rpi is not bricked

(we had to replace radio board to fix this)


lc0001coll (approved):
LGTM! Great fix!

